PR #994, opened 2023-09-21
fix: BX-1030

estebanmino:
Fixes BX-####
Figma link (if any):

announcing provider and set listeners before trying injection


greg-schrammel (approved):

derHowie (approved):
🗣️

PR #997, opened 2023-09-22
Bridge screen

greg-schrammel:
Fixes BX-####
Figma link (if any):

https://github.com/rainbow-me/browser-extension/assets/6232729/1c067f0e-9460-4df8-bb33-4454550adcf5


estebanmino (commented):
image
missing lang

estebanmino (approved):
wdyt about adding an e2e flow for this?

src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
```diff
@@ -52,23 +53,28 @@ export function TokenContextMenu({ children, token }: TokenContextMenuProps) {
...
+
   const explorer = getTokenBlockExplorer(token);
   const hasExplorerLink = !isNativeAsset(token?.address, token?.chainId);
   const isEth = [token.address, token.mainnetAddress].includes(ETH_ADDRESS);
 
   if (isWatchingWallet && !allowSwap && isEth) return <>{children}</>;
 
   return (
-    <DetailsMenuWrapper closed={closed} onOpenChange={onOpenChange}>
+    <DetailsMenuWrapper closed={true} onOpenChange={onOpenChange}>
```
estebanmino, line 77: was this on purpose?

src/entries/popup/pages/swap/SwapTokenInput/TokenDropdown/TokenToBuyDropdown.tsx
```diff
@@ -34,46 +33,44 @@ export const TokenToBuyDropdown = ({
...
+    () => outputChainId && isL2Chain(outputChainId),
+    [outputChainId],
+  );
+
   const assetsCount = useMemo(
     () => assets?.reduce((count, section) => count + section.data.length, 0),
     [assets],
   );
 
+  const t = useTranslationContext();
```
estebanmino, line 46: i feel like this is gonna be painful to maintain in future changes

for example, `explainers.flashbots` doesn't have any difference between swaps and bridges (most of the translations don't), adding that we'll have to maintain all of the languages

why not just change the swap / bridging related text and leave the rest as it was?

i'm imagining having some text like

```
{
  select_tokens `Select tokens to %{action}`
}
```

and that `action` could come from the `useTranslationContext` hook or something like that

wdyt?

src/entries/popup/pages/swap/index.tsx
```diff
@@ -56,6 +56,10 @@ import {
 import { useBrowser } from '../../hooks/useBrowser';
 import useKeyboardAnalytics from '../../hooks/useKeyboardAnalytics';
 import { useKeyboardShortcut } from '../../hooks/useKeyboardShortcut';
+import {
```
estebanmino, line 59: i like that there was no need to copy paste this again for a new bridge route 🚀

estebanmino (commented):
more missing lang

image

estebanmino (commented):
this explainer needs lang updates for bridge 
image

estebanmino (commented):
https://github.com/rainbow-me/browser-extension/assets/12115171/bd334e06-b435-4bcf-9d3e-226266f15b55

seems like flipping is not working correctly

- go to op eth and click bridge
- select mainnet eth
- flip
- press max on mainnet eth, seems like the asset is not correctly pick since it says there's no value 

in the video i go to mainnet eth brige and i see the correct value for the asset and max button works as expected

src/entries/popup/pages/swap/SwapTokenInput/TokenDropdown/TokenToBuyDropdown.tsx
greg-schrammel: yes I was concerned with that too, I'll try this action idea

src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
greg-schrammel: yes, `closed` is kind of a weird prop but I didn't want to refactor it here, in practice it just removes the `onOpenChange` in the underling `ContextMenu`

estebanmino (commented):
looking good, found one more thing

https://github.com/rainbow-me/browser-extension/assets/12115171/2f013ab8-99ec-46b0-84f3-253bed0d70df

we should probably clear swap persistance when going out of the bridge flow? you can see in the video that i'm going to bridge something, then i go to swap and i see the bridge persisted data in the swap flow. Then i continue to other token, go back to swap, and now i see dai <> eth swap persisted data

estebanmino (approved):
looking good

src/core/graphql/queries/metadata.graphql
```diff
@@ -99,11 +100,7 @@ query priceChart(
   }
 }
 
-query dApp(
-  $shortName: String!
-  $url: String!
-  $status: Boolean!
-) {
+query dApp($shortName: String!, $url: String!, $status: Boolean!) {
```
derHowie, line 103: At some point we should figure out what the formatting mismatch is. A lot of files I open now are yellow because you're reformatting with a different config.

derHowie (approved):
lgtm

PR #1000, opened 2023-09-25
pkg bumps

estebanmino:
Fixes BX-####
Figma link (if any):

deep-object-diff is failing on audits so i bumped this and related packages, seems like it's working the same than before


derHowie (approved):
🙏

brunobar79 (approved):
🙏

PR #1001, opened 2023-09-25
Hide unverified token search results for short queries

christianbaroni:

- Speeds up the swap output search by filtering out unverified results if the search query length is less than 3 characters

### DIFF
--- a/src/entries/popup/hooks/useSearchCurrencyLists.ts
+++ b/src/entries/popup/hooks/useSearchCurrencyLists.ts
@@ -400,7 +400,7 @@ export function useSearchCurrencyLists({
     targetVerifiedAssets,
     targetUnverifiedAssets,
     crosschainExactMatches,
-    disableUnverifiedSearch,
+    enableUnverifiedSearch,
   ]);
 
   return {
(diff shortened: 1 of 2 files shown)


derHowie (approved):
👌

greg-schrammel (approved):
isn't it worth it disabling the unverified tokens query too?

christianbaroni:
> isn't it worth it disabling the unverified tokens query too?

@greg-schrammel @derHowie pushed a change for this, mind taking another look?

src/entries/popup/hooks/useSearchCurrencyLists.ts
```diff
@@ -167,6 +168,7 @@ export function useSearchCurrencyLists({
     threshold,
     query,
     fromChainId,
+    disabled: disableUnverifiedSearch,
   });
```
greg-schrammel, line 172: I think useTokenSearch already accepts a config as second arg
```ts
  const {
    data: targetUnverifiedAssets,
    isLoading: targetUnverifiedAssetsLoading,
  } = useTokenSearch(
    {
      chainId: outputChainId,
      keys,
      list: 'highLiquidityAssets',
      threshold,
      query,
      fromChainId,
    },
    { enabled: query.trim().length < 3 },
  );
  ```

estebanmino (approved):

PR #1003, opened 2023-09-25
refactor <AccentColorProvider />

greg-schrammel:

removes `<AccentColorProviderWrapper />` and changes the export definition of the default `<AccentColorProvider />` to handle the edge cases like fallback when color is undefined and poor theme contrast

shouldn't  have any difference anywhere in the app, this should be just a DX change, lmk if otherwise


src/design-system/components/AnimatedRoute/AnimatedRoute.tsx
```diff
@@ -221,7 +218,7 @@ export const AnimatedRoute = forwardRef((props: AnimatedRouteProps, ref) => {
   const containerRef = useRef<HTMLDivElement>(null);
 
   const content = (
-    <AccentColorProviderWrapper
+    <AccentColorProvider
```
estebanmino, line 221: i'd maintain the name `AccentColorProviderWrapper` since `AccentColorProvider` already exists in design-system

estebanmino: actually nvm looks good

estebanmino (approved):

derHowie (approved):
💅

PR #1006, opened 2023-09-26
Fix background coloring in the `/import/select` sheet of onboarding

BrodyHughes:
Fixes [this conversation on slack.](https://rainbowhaus.slack.com/archives/C044WE28BTK/p1695673450423429)
Christian: 'whole background should be `surfaceSecondary`'

| Before | After |
| --- | --- |
| Screenshot 2023-09-26 at 12 52 53 PM | Screenshot 2023-09-26 at 12 47 00 PM |


derHowie (approved):

src/entries/popup/components/FullScreen/FullScreenContainer.tsx
```diff
@@ -2,20 +2,24 @@ import React from 'react';
...
+import { BackgroundColor } from '~/design-system/styles/designTokens';
 
 import { NAVBAR_HEIGHT } from '../Navbar/Navbar';
 
 export function FullScreenContainer({
   children,
+  background,
 }: {
   children: React.ReactNode;
+  background?: string;
```
estebanmino, line 14: i think you could change `string` for `BackgroundColor` here and you can avoid the `background as BackgroundColor` in line 22

estebanmino: and also FullScreenContainer will only accept `BackgroundColor` strings and not random strings so ts can suggest which values to use

estebanmino (approved):

Issue #991, opened 2023-09-20
Improve auditability by removing / replacing dependencies


paulmillr:
Auditing rainbow is complicated, because it's too big. Size can easily be reduced by removing and replacing dependencies.

1. ethers v5 could be replaced with v6 or viem. They both use better cryptography and native bigints instead of bn.js and bignumber.js. Viem is friendly to wagmi, which you are already using.
2. bignumber.js direct dep can be removed in favor of native bigints
3. buffer can be replaced with native uint8array

Probably a deep dive into all deps would reveal hundreds of kilobytes of unused code.

DanielSinclair:
@paulmillr Thanks for having a look! We had actually built much of our keychain before viem and ethers v6, and ethers v5 is still our go-to at Rainbow for a little while yet (and many other projects including MetaMask). A migration to the newer frameworks with better tree-shaking and deprecating BigNumber is definitely something we'll work towards. If you happen to see any other unused or bloated dependencies, definitely let us know; we want to continue to improve cold boot times for short-lived workers in the Manifest v3 world.

paulmillr:
https://github.com/rainbow-me/browser-extension/blob/ef4b4ed75b0963ebbc4a4ec3bf1b4242acfc0e77/src/core/keychain/keychainTypes/keyPairKeychain.ts#L3

This basically says you're using `elliptic` for secp256k1 signatures, which is pretty bad, even though it's sandboxed by lavamoat. Elliptic had 2 CVEs and after that it became basically unmaintained.

The suggestion is to upgrade at least these `Mnemonic` and `HDNode` methods to audited https://github.com/ethereum/js-ethereum-cryptography (used in ethers v6) or directly to scure-bip32 / scure-bip39 which are small and also audited

DanielSinclair:
@paulmillr Thanks for highlighting — we're on it and the team is beginning the upgrade for Ethers v6 this week

PR #1007, opened 2023-09-26
Upgrade cryptography libraries

brunobar79:
Fixes BX-####
Figma link (if any):

While this is not an ideal approach, it's the shortest path to upgrade all of our cryptography operations.
As reported via #991,  there are now better alternatives to ethers v5 (which uses elliptic) so this is the main motivation behind it.

We have considered the following approaches:
- Migrating to Ethers v6 -  It's not an easy task, doable but long and one of the main issues is its incompability with Wagmi < 1 and even after that we'd need an adapter
- MIgrating to Viem - Perf. wise seems like the best path forward.  Similar effort maybe a bit less but still a lot of work. This also requires a full upgrade of our swaps SDK since it's also using ethers v5

Both of those approaches are better in the long term (less dependencies, usage of new native browser primitives, smaller bundle size, etc) but that needs to be planned appropriately so this is a temporary security upgrade to use the best cryptography libraries available at the lowest level ( mnemonic / wallet generation and signing) ASAP

## What to test

- Onboarding via pkey
- Onboarding via seedphrase
- Send
- Swap + approvals (with permit - ex. USDC to ANY and without permit - ex. WETH to any token)
- Personal sign
- Sign Typed Data


BrodyHughes (approved):
Checked all these items:
- Onboarding via pkey
- Onboarding via seedphrase
- Send
- Swap + approvals (with permit - ex. USDC to ANY and without permit - ex. WETH to any token)
- Personal sign
- Sign Typed Data
All look good

src/core/keychain/RainbowSigner.ts
```diff
@@ -0,0 +1,93 @@
...
+    // are made using EthereumJS instead of ethers v5
+
+    const typedTx = TransactionFactory.fromTxData({
+      data: transaction.data?.toString(),
+      to: transaction.to,
+      accessList: [],
+      maxPriorityFeePerGas: transaction.maxPriorityFeePerGas
+        ? BigNumber.from(transaction.maxPriorityFeePerGas).toHexString()
+        : undefined,
+      maxFeePerGas: transaction.maxFeePerGas
```
estebanmino, line 59: no gasPrice for legacy supported chains?

src/core/keychain/keychainTypes/hdKeychain.ts
```diff
@@ -87,9 +99,10 @@ export class HdKeychain implements IKeychain {
 
   getSigner(address: Address): Signer {
     const _privates = privates.get(this)!;
-    const wallet = _privates.getWalletForAddress(address);
-    if (!wallet?._isSigner) throw new Error('Not a signer');
-    return wallet;
+    const provider = getProvider({ chainId: mainnet.id });
```
estebanmino, line 102: i checked l2 sends because of this, working fine, but chainId as mainnet doesn't do anything?

brunobar79: The interface of signer requires to pass a provider to the constructor. That's why it's set to mainnet but when used through Wallet we always pass the right provider

src/core/keychain/RainbowSigner.ts
brunobar79: good call. missed that one. Just added it

package.json
```diff
@@ -113,13 +116,15 @@
     "@vanilla-extract/sprinkles": "1.5.0",
     "@wagmi/chains": "1.6.0",
     "bignumber.js": "9.0.1",
+    "bip39": "3.1.0",
```
paulmillr, line 119: I believe this library is unused

paulmillr:
looking good other than the unused dep

estebanmino (approved):

package.json
brunobar79: you're right! thanks for the review 🙏

PR #1008, opened 2023-09-26
e2e: shortcut onboarding helper

BrodyHughes:
adds a helper for onboarding using all shortcuts / keyboard nav. will use this in the shortcut tests.


estebanmino (commented):
are we using this somewhere?

BrodyHughes:
> are we using this somewhere? 

nope! but i'm going to in the next tests i'm writing. just was trying to keep the PRs smaller / simpler.

estebanmino:
> > are we using this somewhere?
> 
> @estebanmino nope! but i'm going to in the next tests i'm writing. just was trying to keep the PRs smaller / simpler.

wdyt to mrge this when the first test you use this with? imo we shouldn't merge code we're not using

BrodyHughes:
adding this to the new test i wrote

PR #1009, opened 2023-09-26
feat: prefetch dapp metadata

estebanmino:
Fixes BX-####
Figma link (if any):

prefetch dapp metadata so when the user gets a popup request the dapp metadata (names and image) is already loaded


derHowie (approved):
⏩

brunobar79 (approved):

PR #1018, opened 2023-09-28
fix submenus

greg-schrammel:
Fixes BX-####
Figma link (if any):

before / after

https://github.com/rainbow-me/browser-extension/assets/6232729/a26ea187-fb58-47e1-af31-e750dd15d66e

https://github.com/rainbow-me/browser-extension/assets/6232729/c4c7f0c5-69f6-43cd-a9c6-34257a7897d8


estebanmino (changes requested):
the submenu is centered but it shouldn’t scale down, it should maintain the same width than before the animation

EDIT: this is not an issue anymore

src/entries/popup/components/DropdownMenu/DropdownSubMenu.tsx
```diff
@@ -176,16 +176,6 @@ export const DropdownSubMenu = ({
       <DropdownMenuTrigger asChild>
         <Box ref={subMenuElementRef}>{subMenuElement}</Box>
       </DropdownMenuTrigger>
-      <DropdownMenuContent
```
estebanmino, line 179: this shouldn't be removed

greg-schrammel: what is this for?

estebanmino (approved):
🪨

derHowie (approved):

PR #1029, opened 2023-10-03
[BX-1053] Tokens should be default tab selection after switching wallets

derHowie:
## What to test
Make sure you land on the tokens tab after selecting another address


estebanmino (approved):
this is working fine but should we add it to switching wallets via shortcuts?

greg-schrammel (approved):

PR #1030, opened 2023-10-03
[BX-1057] ‘Del’ text slightly exceeds button width

derHowie:
Let multicharacter button strings expand the width of the bubble keys

Screen Shot 2023-10-03 at 7 14 46 AM

## What to test
Make sure button strings have sufficient horizontal padding inside the bubble keys


estebanmino (commented):
i feel like we should only do this when the shortcut has more than one character

for example

image


W is bigger than other letters so in the app menu these shortcut hints look misaligned

wdyt?

derHowie:
> i feel like we should only do this when the shortcut has more than one character
> 
> for example
> 
> image
> W is bigger than other letters so in the app menu these shortcut hints look misaligned
> 
> wdyt?

Good call, fixed.

Screen Shot 2023-10-04 at 12 41 12 AM

greg-schrammel (approved):

estebanmino:
> > i feel like we should only do this when the shortcut has more than one character
> > for example
> > image
> > W is bigger than other letters so in the app menu these shortcut hints look misaligned
> > wdyt?
> 
> Good call, fixed.
> 
> Screen Shot 2023-10-04 at 12 41 12 AM

hm it doesn't look fixed 

image

src/entries/popup/components/ShortcutHint/ShortcutHint.tsx
```diff
@@ -11,13 +11,14 @@ export const ShortcutHint = ({
 }) => {
   const height = small ? '14px' : '18px';
   const width = small ? '16px' : '18px';
+  const widthKey = hint?.length || 0 > 1 ? 'minWidth' : 'width';
```
estebanmino, line 14:
```suggestion
  const widthKey = (hint?.length || 0) > 1 ? 'minWidth' : 'width';
```

brunobar79 (approved):
🤝

PR #1032, opened 2023-10-03
fix: BX-1046

estebanmino:
Fixes BX-1046
Figma link (if any):

- now using debounce for useOptimismL1SecurityFee
- now using debounce for useEstimateGasLimit
- calculateL1FeeOptimism was reassigning transactionRequest object so all hooks dependent on it were retriggering a lot of times, like the two hooks i mentioned above


brunobar79 (approved):
💯

derHowie (approved):

PR #1036, opened 2023-10-04
Sentry filtering beforeSend

brunobar79:
Fixes BX-####
Figma link (if any):

There's certain events that we don't want to send to Sentry at all because they either can't be fixed or happen way to often (or both in this case)  so we're filtering those out to prevent an impact in our quota.


greg-schrammel (approved):

estebanmino (approved):

src/core/sentry/index.ts
```diff
@@ -3,6 +3,15 @@ import { BrowserTracing } from '@sentry/tracing';
 
 import { version } from '../../../package.json';
 
+// Any error that we don't wanna send to sentry should be added here
+// via partial match
+const IGNORED_ERRORS = [
+  'Duplicate script ID',
```
estebanmino, line 9: this one was fixed

brunobar79: SInce it's #1 o #2 It won't hurt keep in it here in case it comes back

PR #1040, opened 2023-10-04
Add a maxHeight to externalImage component

BrodyHughes:
The bottom of the NFT images has a slight gap. The external Image component had a slightly larger height than expected. This should fix it. 
| Before | After |
|--------|-------|
| Screenshot 2023-10-04 at 3 03 12 PM | Screenshot 2023-10-04 at 3 34 44 PM |
| Screenshot 2023-10-04 at 3 35 00 PM | Screenshot 2023-10-04 at 3 40 40 PM |


src/entries/popup/components/CoinIcon/CoinIcon.tsx
```diff
@@ -282,12 +282,13 @@ export const NFTIcon = ({
         width={size}
         borderRadius={nftRadiusBySize[size]}
         onError={() => setBadSrc(true)}
+        maxHeight={size}
       />
-      <Box position="absolute" bottom="0" style={{ zIndex: 2, left: '-6px' }}>
-        {badge && chainId !== ChainId.mainnet && (
+      {badge && chainId !== ChainId.mainnet && (
```
BrodyHughes, line 287: also moved this to not render this empty box for the NFT components

greg-schrammel: nicee

src/entries/popup/components/ExternalImage/ExternalImage.tsx
```diff
@@ -60,7 +61,13 @@ const ExternalImage = (
 
   if (!signedUrl) return null;
   return (
-    <Box style={{ overflow: 'clip' }} borderRadius={props.borderRadius}>
+    <Box
+      style={{
+        overflow: 'clip',
+        maxHeight: props.maxHeight ? props.maxHeight : '',
```
greg-schrammel, line 67: I think we can safely set `maxHeight` to `height`, so we don't need the extra `maxHeight` prop

BrodyHughes: ok cool. wasn't sure if this was a good idea or not / had other implications elsewhere that i hadn't considered.

BrodyHughes: fixed.

brunobar79 (approved):
💯

greg-schrammel (approved):

PR #1058, opened 2023-10-13
RPC proxy

brunobar79:
Fixes BX-940
Figma link (if any):

Shift public provider RPC calls through our Backend proxy. This is a stepping stone towards RPC Defaults selection (i.e. users could choose an RPC like [MEV Blocker](https://mevblocker.io/) from our allowlist) and Custom Networks. This will help us work around the Manifest v3 limitations. 

The feature will only work if we switch this new env var `RPC_PROXY_ENABLED` to true.

Requires: https://github.com/rainbow-me/browser-extension-env/pull/14


derHowie (approved):
👌

src/core/wagmi/createWagmiClient.ts
```diff
@@ -21,47 +22,55 @@ const noopStorage = {
   removeItem: () => null,
 };
 
+const getOriginalRpcEndpoint = (chain: Chain) => {
+  switch (chain.id) {
+    case ChainId.hardhat:
+      return { http: chain.rpcUrls.default.http[0] };
+    case ChainId.mainnet:
+      return { http: process.env.ETH_MAINNET_RPC as string };
```
greg-schrammel, line 30: can't you type it as string in `.env.d.ts`? not sure how it works

greg-schrammel (approved):

src/core/wagmi/createWagmiClient.ts
brunobar79: Good call!

PR #1067, opened 2023-10-17
useNfts initial

derHowie:
added calls to simplehash nft proxy and created `useNfts`

## What to test
None of this code should be in use, but let's just make sure this addition doesn't affect the apps behavior.


src/core/network/nfts.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const fetchPolygonAllowList =
+  async (): Promise<PolygonAllowListDictionary> => {
+    const allowList = await rainbowFetch<{ data: { addresses: string[] } }>(
+      'https://metadata.p.rainbow.me/token-list/137-allowlist.json',
+      { method: 'get' },
+    );
+    const polygonAllowListDictionary = allowList.data?.data?.addresses?.reduce(
+      (allowListDict, tokenAddress) => {
+        allowListDict[tokenAddress] = true;
```
greg-schrammel, line 112: just to make sure the api return `tokenAddress` all lowercase and not checksummed, as that would break the filter later

greg-schrammel (approved):
👨‍🍳🤌

src/core/network/nfts.ts
```diff
@@ -0,0 +1,118 @@
...
+      message: (e as Error)?.message,
+    });
+    return [];
+  }
+};
+
+export const fetchPolygonAllowList =
+  async (): Promise<PolygonAllowListDictionary> => {
+    const allowList = await rainbowFetch<{ data: { addresses: string[] } }>(
+      'https://metadata.p.rainbow.me/token-list/137-allowlist.json',
```
estebanmino, line 107: why this doesn't have a `RainbowFetchClient` like all other requests?

src/core/resources/nfts/nfts.ts
```diff
@@ -0,0 +1,114 @@
...
+    40,
+  );
+  const nftRequests = collectionsBatches.map((collectionIds) =>
+    fetchNfts({
+      address,
+      chains,
+      collectionIds,
+    }),
+  );
+  const nftsResponse = (await Promise.all(nftRequests)).flat();
```
estebanmino, line 86: could be possible that one of these fetches fail and we end up with no nfts for any chain?

greg-schrammel: good point, `promise.allSettled` may be best here

src/core/network/nfts.ts
derHowie: No reason. I created an http client for this based on your comment.

src/core/network/nfts.ts
estebanmino: are we good here?

estebanmino (approved):
lgtm

estebanmino:
commenting about this jic https://github.com/rainbow-me/browser-extension/pull/1067#discussion_r1371865927

PR #1079, opened 2023-10-23
fix: do not default to mm provider when toggle is off

estebanmino:
Fixes BX-1088
Figma link (if any):

now:

when our default rnbw provider toggle is off and the user has metamask, we're relaying requests to mm provider
this was probably causing the issue, users had the tooggle off, eip6963 was picking rainbow correctly but we were relaying requests to mm

with this pr:
when our default rnbw provider toggle is off and the user has metamask we're still using rainbow provider if the dapp pick the rainbow provider 
this won't affect the window.ethereum override when the rnbw by default tooggle is on since that logic is happening on `inpage`

### DIFF
--- a/src/core/providers/RainbowInjectedProvider.ts
+++ b/src/core/providers/RainbowInjectedProvider.ts
@@ -26,13 +26,6 @@ export type RequestResponse =
       result: any;
     };
 
-const getMetaMaskProvider = () => {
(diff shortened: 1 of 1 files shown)


derHowie (approved):
👌

greg-schrammel (approved):
gotha! nice

PR #1090, opened 2023-10-26
feat: custom rpc settings

estebanmino:
Fixes BX-####
Figma link (if any):

adding custom rpc settings
custom rpc state
ignore the bad ui, pushing it over a feature flag while we wait for designs

https://www.loom.com/share/d72a2ec0efb84704ae312c71020c86e2


src/entries/popup/pages/settings/customRpc.tsx
```diff
@@ -0,0 +1,232 @@
...
+import { maskInput } from '../../components/InputMask/utils';
+
+export function SettingsNetworksCustomRPC() {
+  const { customRPCs, addCustomRPC } = useCustomRPCsStore();
+  const [customRPC, setCustomRPC] = useState<{
+    rpcUrl?: string;
+    chainId?: number;
+    name?: string;
+    symbol?: string;
+    explorerUrl?: string;
```
brunobar79, line 17: Should we add another field `explorerName` ? We could use it everywhere where we say stuff like "view on Etherscan"

brunobar79, line 17: also, can we add another flag "active" ? that way we know which one to use it in case you have multiple rpcs for the same chain Id

brunobar79 (approved):
🤝 🤝 🤝

src/entries/popup/pages/settings/customRpc.tsx
estebanmino: oh right

greg-schrammel: I think we're using the url hostname rn, (I kinda like that rly) but should change so that we grab from the rpc config instead of wagmi 
Screenshot 2023-10-27 at 13 55 57

greg-schrammel (approved):
👌

PR #1093, opened 2023-10-26
fix e2e

BrodyHughes:
replace findText instance with findByTestId. Unsure why it's not working but this is a better more reliable method anyways.


brunobar79 (approved):
🔥

derHowie (approved):

PR #1097, opened 2023-10-27
[BX-1084] feat: export public addresses

BrodyHughes:
**Changes:**
This adds a shortcut thats accessible via the magic menu (`cmd + K`) that exports all wallets into a CSV file with the headers `public_address`, `name`, and `type`. Data is pulled from:
- `public_address` is pulled via `getWallets() > accounts[i]`, 
- `name` is pulled via `const { walletNames } = useWalletNamesStore()`
- `type` is pulled via `getWallets() > type` and assigned `Watching` if its `ReadOnlyKeychain`, `Imported` if it's `HdKeychain`, and `Hardware Wallet - ${vendor}` if it's `HardwareWalletKeychain`

Daniel wants to add a Unique device number to each HW that will be added later to another column. This requires some keychain changes so i want to keep it out of this PR to keep the scope more manageable for me.

**Flow in action:**
https://github.com/rainbow-me/browser-extension/assets/41711440/9a51c0f2-2f90-40b7-9133-63f5931db387


static/manifest.json
```diff
@@ -59,6 +59,7 @@
   "minimum_chrome_version": "88",
   "name": "Rainbow DEVELOPMENT BUILD",
   "permissions": [
+    "downloads",
```
DanielSinclair, line 62: This might be an issue because it could require an ugly permission approval on the next update; something to ask Bruno about.

src/entries/popup/hooks/useHomeShortcuts.ts
```diff
@@ -73,6 +76,57 @@ export function useHomeShortcuts() {
...
+        url: url,
+        filename: 'accounts.csv',
+      },
+      () => {
+        URL.revokeObjectURL(url);
+      },
+    );
+    triggerToast({
+      title: 'Address CSV file created',
+      description: 'downloading...',
```
DanielSinclair, line 126: "Addresses Downloaded" with the CSV file name as the description would be nice here; just like the Copied toast

src/entries/popup/hooks/useHomeShortcuts.ts
```diff
@@ -73,6 +76,57 @@ export function useHomeShortcuts() {
...
+    const data = await getWallets();
+    console.log('getWallets', data);
+    const csvContent = generateCSV(data as WalletObj[]);
+    const blob = new Blob([csvContent], { type: 'text/csv' });
+    const url = URL.createObjectURL(blob);
+
+    chrome.downloads.download(
+      {
+        url: url,
+        filename: 'accounts.csv',
```
DanielSinclair, line 118: Let's call it something like `rainbow_addresses.csv` which is similar to what MetaMask does with `MetaMask state logs.json`

src/entries/popup/hooks/useHomeShortcuts.ts
```diff
@@ -73,6 +76,57 @@ export function useHomeShortcuts() {
...
+    (data: WalletObj[]) => {
+      let csvContent = 'public_address,name,type\n';
+
+      data.forEach(
+        (item: { accounts: string[]; type: string; vendor?: string }) => {
+          const address = item.accounts[0];
+          // If address exists in walletNames, use that name. Otherwise, use defaultName.
+          const name = walletNames[address as Address] || item.accounts[0];
+          const importedType =
+            item.type === 'ReadOnlyKeychain' ? 'Watching' : 'Imported';
```
DanielSinclair, line 95: I think we'll need to account for hardware wallets here as well

static/manifest.json
BrodyHughes: @brunobar79 any thoughts here?

src/entries/popup/components/CommandK/useCommands.tsx
```diff
@@ -229,6 +231,15 @@ export const staticCommandInfo: CommandInfo = {
     symbolSize: 14,
     type: SearchItemType.Shortcut,
   },
+  exportAddressList: {
+    name: 'Export address list to CSV',
```
DanielSinclair, line 235: I like the CSV addition. What about "Export Addresses as CSV" instead for capitalization consistency?

DanielSinclair: I realized these strings aren't covered by i18n, so created a backlog ticket for that here: https://linear.app/rainbow/issue/BX-1112/command-k-command-name-i18n

src/entries/popup/components/CommandK/utils.ts
```diff
@@ -427,3 +430,61 @@ export function useKeyboardNavigation(
...
+  let csvContent = 'public_address,name,type\n';
+
+  data.forEach(({ accounts, type, vendor }) => {
+    const address = accounts[0];
+    const name = walletNames[address as Address] || address;
+
+    let importedType = typeMapping[type];
+
+    if (type === 'HardwareWalletKeychain') {
+      importedType = `Hardware Wallet - ${vendor}`;
```
DanielSinclair, line 462: To deobfuscate a user with multiple hardware devices, what if keep this as `Hardware Wallet` and then add an additional column called `device` that would have the manufacturer and the device ID (i.e. #1) so `Ledger #1` and `Ledger #2` and `Trezor #1`. IIRC those number ids / UUID should be available in the keychain

DanielSinclair: Otherwise working well and looking good. Nice icon picks 👌

static/manifest.json
BrodyHughes: moving away from permission changes by doing this:
```
  const data = await getWallets();
  const csvContent = generateCSV(data as WalletObj[], walletNames);
  const blob = new Blob([csvContent], { type: 'text/csv' });
  const url = URL.createObjectURL(blob);

  // Create a temporary anchor to initiate the download
  const tempLink = document.createElement('a');
  tempLink.href = url;
  tempLink.download = 'rainbow_addresses.csv';
  document.body.appendChild(tempLink);
  tempLink.click();
  document.body.removeChild(tempLink);

  // Release the object URL after use
  URL.revokeObjectURL(url);
  ```

File: src/entries/popup/components/CommandK/utils.ts (hunk 6 above)
Comment (BrodyHughes) [reply src/entries/popup/components/CommandK/utils.ts:462]: This i think requires a few bigger changes to keychain things so will look at this in a followup PR to keep scope here reasonable.

File: src/entries/popup/components/CommandK/utils.ts [hunk 7]
```diff
@@ -427,3 +431,53 @@ export function useKeyboardNavigation(
...
+};
+
+export const generateCSV = (
+  data: KeychainWallet[],
+  walletNames: WalletNames,
+): string => {
+  let csvContent = 'public_address,name,type\n';
+
+  data.forEach(({ accounts, type, vendor }) => {
+    const address = accounts[0];
```
greg-schrammel, line 451: if it's a seed phrase with multiple derived addresses I think this will make it only output the first one (?)

BrodyHughes: seems true. hmmmmmmmm

DanielSinclair: Yep, confirmed this issue

DanielSinclair:
UI and download flow is looking good 👌

src/entries/popup/components/CommandK/utils.ts
BrodyHughes: @DanielSinclair @greg-schrammel 
Ok, fixed this functionality, and added a column indicating which wallet is the parent as well.

Screenshot 2023-11-02 at 9 53 51 AM

DanielSinclair (approved):
Tested with a watched wallet, and 2 wallet groups

src/entries/popup/components/CommandK/utils.ts
DanielSinclair: @greg-schrammel Tweak it a bit, but the original problem and private keys should be handled now. Should be ready for merge!

src/entries/popup/components/CommandK/utils.ts
```diff
@@ -427,3 +431,50 @@ export function useKeyboardNavigation(
...
+  const tempLink = document.createElement('a');
+  tempLink.href = url;
+  tempLink.download = 'rainbow_addresses.csv';
+  document.body.appendChild(tempLink);
+  tempLink.click();
+  document.body.removeChild(tempLink);
+  URL.revokeObjectURL(url);
+
+  triggerToast({
+    title: 'Addresses Downloaded',
```
estebanmino, line 477: no internacionalization?

src/entries/popup/components/CommandK/utils.ts
```diff
@@ -427,3 +431,50 @@ export function useKeyboardNavigation(
     };
   }, []);
 }
+
+const typeMapping: { [key: string]: string } = {
+  [KeychainType.ReadOnlyKeychain]: 'Watching',
+  [KeychainType.HdKeychain]: 'Recovery Phrase',
+  [KeychainType.KeyPairKeychain]: 'Private Key',
+  [KeychainType.HardwareWalletKeychain]: 'Hardware Wallet',
```
estebanmino, line 439: no internacionalization?

estebanmino (approved):
minor comments about lang files but lgtm

PR #1102, opened 2023-10-30
Storage refactor

brunobar79:
Fixes BX-####
Figma link (if any):

did some research about SessionStorage quota and [there's actually a 10MB limit](https://developer.chrome.com/docs/extensions/reference/storage/#property-session) but for a normal user we store less than 20kb.

However there's 2 keys that can grow unlimitedly during the session which are:
- rateLimit (per dapp) - this stores a key for the host and a timestamp for every dapp that you have connected to in the current browser's session
- queuedEvents:  this are all the tracking event from the service worker that we put in memory to be able to send it to the server through the popup (since we can't do it directly from the service worker)

I highly doubt any of these two is a real offender but anyways I'm taking two actions:
- Replaced our direct usage of chrome.storage.session and chrome.storage.local for our own wrappers (LocalStorage and Session Storage).
This allow us to improve the usage a big (chrome api sucks), we can add custom error handlers or process the data before storing it (encryption, compression, etc).
- Added logging in case the quota is exceeded so we can see what's going on.

Gonna archive the sentry error and keep an eye in the future for the new errors that we might get.

## What to test

Passing tests should be enough

### DIFF
--- a/src/core/storage/index.ts
+++ b/src/core/storage/index.ts
@@ -1,4 +1,4 @@
-export const Storage = {
+export const LocalStorage = {
   async clear() {
     await chrome?.storage?.local?.clear();
   },
--- a/src/core/storage/index.ts
+++ b/src/core/storage/index.ts
@@ -1,3 +1,5 @@
+import { RainbowError, logger } from '~/logger';
+
 export const LocalStorage = {
   async clear() {
     await chrome?.storage?.local?.clear();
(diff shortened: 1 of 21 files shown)


src/core/storage/index.ts
```diff
@@ -29,3 +31,64 @@ export const Storage = {
...
+
+        const queuedEvents = await SessionStorage.get('queuedEvents');
+        const rateLimits = await SessionStorage.get('rateLimits');
+
+        logger.info(
+          'SessionStorage queuedEvents size: ',
+          queuedEvents?.length || 0,
+        );
+        logger.info(
+          'SessionStorage queuedEvents size: ',
```
estebanmino, line 62: same key than the log above and for different info

estebanmino (approved):
lgtm

derHowie (approved):
👌

PR #1104, opened 2023-10-31
fix: BX-1111

estebanmino:
Fixes BX-1111
Figma link (if any):

rate limit was blocking the provider request

with this pr we're now only blocking requests that need a provider request


brunobar79 (approved):
🤝

derHowie (approved):

PR #1111, opened 2023-11-03
feat: support no price and no time for custom network estimated fees

estebanmino:
Fixes BX-####
Figma link (if any):

## What changed (plus any additional context for devs)

for custom networks where we don't have the native asset price, estimated fee will start showing up as crypto value
we will not show estimated times for these cases as well

PR #1123, opened 2023-11-08
Improve custom chain support

brunobar79:
Fixes BX-####
Figma link (if any):

- We now have asset prices, charts for all the chains that we unofficially support through the backend.
- We hide prices for chains that we don't have any info
- We have asset & chain icons for a lot of popular chains taken from the assets repo.

http://recordit.co/5rM9BOLzol


src/core/resources/assets/customNetworkAssets.ts
```diff
@@ -110,42 +140,85 @@ async function customNetworkAssetsFunction({
...
+              asset: customNetworkNativeAssetParsed,
+              currency,
+              balance: nativeAssetBalance.toString(),
+            });
 
-            parsedAssetsDict[chain.id as ChainId] = {
-              [customNetworkNativeAssetParsed.uniqueId]:
-                customNetworkNativeAssetParsed,
-            };
+            console.log(
```
estebanmino, line 204: log

src/entries/popup/components/TransactionFee/TransactionFee.tsx
```diff
@@ -168,6 +201,12 @@ function Fee({
                   <TextOverflow weight="semibold" color="label" size="14pt">
                     {isLoading
                       ? '~'
+                      : !nativeCurrencyGasFeeSupported(chainId)
+                      ? convertToNativeGasFee(
```
estebanmino, line 205: if the intention of this was using the fee as native value if the price is not available that should have been done here https://github.com/rainbow-me/browser-extension/pull/1111

that should be checking if native asset has price and if not default to crypto values

estebanmino: to fix it you can change https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/hooks/useGas.ts#L54C23-L54C47

`useNativeAssetForNetwork` to `useNativeAsset`

estebanmino (approved):

derHowie (approved):
lgtm

PR #1124, opened 2023-11-08
fix: asset filtering by networks settings

estebanmino:
Fixes BX-####
Figma link (if any):

filtering assets by network broke, so any change on networks settings weren't filtering assets correctly, this pr fixes that


greg-schrammel (approved):
nice

derHowie (approved):

PR #1129, opened 2023-11-13
[BX-1098] NFT Infinite Query

derHowie:
Gradually grabbing nfts via infinite query and implemented NFT_LIMIT. This NFT_LIMIT is arbitrary. In the future we can play with increasing the value.

## What to test
Ensure that the requests are working and nfts are gradually showing up in chunks. Check a huge wallet to make sure the limit prevents the app from experiencing performance issues.


brunobar79 (approved):
💯

greg-schrammel (approved):

PR #1141, opened 2023-11-16
walllet_WatchAsset

brunobar79:
Fixes BX-1075
Figma link (if any):

- Adds support for `wallet_watchAsset`
- First we try to use the session chain Id if there's one, if not, we look at the host and we know the block explorers for some networks. If we can't figure it out we default to mainnet and the user can switch it. See link below.

https://recordit.co/LakhR5U4Q3

### DIFF
--- a/src/core/utils/chains.ts
+++ b/src/core/utils/chains.ts
@@ -17,6 +17,8 @@ import { isLowerCaseMatch } from './strings';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 export const customChainIdsToAssetNames: Record<ChainId, string> = {
+  1: 'ethereum',
+  10: 'optimism',
   43114: 'avalanchex',
   100: 'xdai',
   324: 'zksync',
--- a/src/core/utils/chains.ts
+++ b/src/core/utils/chains.ts
@@ -17,8 +17,6 @@ import { isLowerCaseMatch } from './strings';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 export const customChainIdsToAssetNames: Record<ChainId, string> = {
-  1: 'ethereum',
-  10: 'optimism',
   43114: 'avalanchex',
   100: 'xdai',
   324: 'zksync',
--- a/src/core/utils/chains.ts
+++ b/src/core/utils/chains.ts
@@ -177,3 +177,36 @@ export const getChainMetadataRPCUrl = async ({
   }
   return null;
 };
+
+export const deriveChainIdByHostname = (hostname: string) => {
+  switch (hostname) {
+    case 'etherscan.io':
+      return ChainId.mainnet;
+    case 'goerli.etherscan.io':
+      return ChainId.goerli;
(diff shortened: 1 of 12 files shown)


src/core/utils/chains.ts
```diff
@@ -163,3 +163,36 @@ export const chainIdToUse = (
...
+    case 'ftmscan.com':
+      return 250;
+    case 'explorer.celo.org':
+      return 42220;
+    case 'explorer.harmony.one':
+      return 1666600000;
+    case 'explorer.avax.network':
+    case 'subnets.avax.network':
+    case 'snowtrace.io':
+      return 43114;
```
estebanmino, line 194: you could use `avalanche.id` from wagmi/chains for this one, and probably for the one above to get the chainid from const

src/entries/background/handlers/handleProviderRequest.ts
```diff
@@ -393,6 +398,57 @@ export const handleProviderRequest = ({
...
+            if (!address) {
+              throw new Error('Address is required');
+            }
+
+            const activeSession = getActiveSession({ host });
+            let chainId = null;
+            if (activeSession) {
+              chainId = activeSession?.chainId;
+            } else {
+              chainId = deriveChainIdByHostname(host);
```
estebanmino, line 431: we should probably block this request if there's no session because if you don't have a session on a network your trying to add a token manually means that you don't have the custom rpc added

this case will produce that `getProvider` calls you have below in the ui will crash the ui

estebanmino: actually no, you could have the rpc with no session, but this is still a problem when the user doesn't have a session but the user also doesn't have the rpc

estebanmino: i believe it defaults to mainnet now that i remember, so it won't crash but will show wrong info to the user?

brunobar79: We discussed this on slack.  Dapps like etherscan have no session to suggest a token so we can't block it.
The user having the RPC or not is a separate thing.  A token can be suggested for mainnet or any chain we do support.  If we don't know the chain we default to mainnet and show a warning,  but we let the user choose the network from the list. There's a screen recording on the "screenshot" section of the PR.

src/entries/popup/pages/messages/WatchAsset/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+      if (!symbol) {
+        const tokenSymbol = await tokenContract.symbol();
+        assetWithMetadata.symbol = tokenSymbol;
+      }
+      const tokenName = await tokenContract.name();
+      assetWithMetadata.name = tokenName;
+      assetWithMetadata.chainId = Number(selectedChainId);
+      assetWithMetadata.chainName = (getChain({
+        chainId: Number(selectedChainId),
+      }).name || '') as ChainName;
```
estebanmino, line 107: you can use `useAssetMetadata` instead to get asset metadata

estebanmino (approved):

derHowie (approved):

PR #1144, opened 2023-11-17
feat: custom rpc form validations

estebanmino:
Fixes BX-####
Figma link (if any):

- added `Form` and `FormInput` components
- when adding a custom rpc, we'll check if the url is a valid rpc endpoint getting the chainId and presenting it to the user in the chainId input of the form
- validations for custom chain was added and confirmed is working
- - when adding a custom asset, we'll check if the address is a valid address and pull the asset metadata
- validations for custom asset was added and confirmed is working

https://www.loom.com/share/95a8b1c89618435ca85749e2b936a173

### DIFF
--- a/src/core/resources/chains/chainMetadata.ts
+++ b/src/core/resources/chains/chainMetadata.ts
@@ -1,0 +1,69 @@
+import { useQuery } from '@tanstack/react-query';
+
+import {
+  QueryConfig,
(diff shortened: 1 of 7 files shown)


brunobar79 (approved):
🤝 🤝 🤝

greg-schrammel (approved):

PR #1146, opened 2023-11-20
fix: BX-1090

estebanmino:
Fixes BX-####
Figma link (if any):

couldn't really repro the ticket so i added a crash safe value for appName


greg-schrammel:
I believe to have fixed it in the simulation pr

brunobar79 (approved):

PR #1148, opened 2023-11-21
remove accept request auto focus

greg-schrammel:
Fixes BX-1125
Figma link (if any):

Deprecate the automatic highlight of Sign, now that we have the Cmd+Return feature


estebanmino (approved):

DanielSinclair:
Nice! 👌 

Noticed 1 issue where the first tab press on Sign Typed Data and Sign Message seems to be non reactive. Maybe the new Copy button takes that tab index, or there is an off by 1 issue somewhere? Likewise for the last tab after the Data tab is active on the Send Transaction prompt.

Also, on the Send Transaction prompt, we had legacy behavior where the Cancel button would be auto-selected. To keep these aligned, what if we remove that default highlight. `ESC` feels more intuitive for the destructive action, and I can imagine a scenario where the default `return` behavior would make the user think they accidentally just sent the transaction (especially for dapps with bad handling of that). Whereas now they'd first have to press tab to cancel, then tab again to send, or simply cmd+return to send faster. Open to your thoughts!

PR #1151, opened 2023-11-23
Fix popup positioning

christianbaroni:
Fixes BX-872

- Fixes the positioning of the popup window that appears when connecting or signing — should now correctly position to the top right of the active browser window
- Adjusts the popup height determination to take into account the estimated size of the popup window's title bar, based on the `userAgent`

Screenshot 2023-11-23 at 3 01 01 PM

## What to test
- When connecting or signing, the popup window should be positioned to the top right of the active browser window
- Popups should now appear on the correct screen if external monitors are being used
- The popup window should be sized correctly and should not be scrollable, at least on Macs


greg-schrammel (approved):
looks good
I think arc saves the popup dimensions how you last arranged it, but chrome is right
didn't test with an external monitor

estebanmino (approved):

DanielSinclair:
Working perfectly on my display link monitors. Nice 👌

PR #1155, opened 2023-11-27
[BX-1106] Skeleton / Loading State Nft Gallery

derHowie:
I added skeleton loaders on the NFT Gallery page. They differ depending on the display mode selected.

PoW: https://recordit.co/xZ3LorqoNe

## What to test
Make sure the skeleton loaders appear when `isInitialLoading` is true.

### DIFF
--- a/src/entries/popup/pages/home/NFTs/NFTs.tsx
+++ b/src/entries/popup/pages/home/NFTs/NFTs.tsx
@@ -30,6 +30,7 @@ import {
 } from '~/design-system';
 import { useContainerRef } from '~/design-system/components/AnimatedRoute/AnimatedRoute';
 import { Lens } from '~/design-system/components/Lens/Lens';
+import { Skeleton } from '~/design-system/components/Skeleton/Skeleton';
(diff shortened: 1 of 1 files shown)


estebanmino (approved):

src/entries/popup/pages/home/NFTs/NFTs.tsx
```diff
@@ -172,96 +168,185 @@ export function PostReleaseNFTs() {
...
+      style={{
+        display: 'flex',
+        flexWrap: 'wrap',
+        flexDirection: 'row',
+        justifyContent: 'flex-start',
+        gap: 16,
+        paddingBottom: 16,
+      }}
+    >
+      {Array.from({ length: 9 }).map((_, i) => {
```
brunobar79, line 295: Idk if the 9 is also used on the actual UI (not skeleton) but if that's the case maybe we could export and reuse on both cases.

Same for the 16 that on the collections a few lines below

brunobar79 (approved):

PR #1166, opened 2023-11-30
[BX-1102] Ens Registration Info

derHowie:
Figma link (if any): https://www.figma.com/file/hzfwKOddfRmeVO4KUsG9wM/Rainbow-BX?type=design&node-id=8127-263102&mode=design&t=P36zv51ZbKE5GZDM-0

The nft details page now displays registration info in place of the price section.

PoW: https://recordit.co/uFOZyEB6aK

## What to test
Visit NFT details for ENS names, then make sure "Registered on" and "Expires in" displays correctly.

For documentation's sake: this is part two of BX-1102. The profile info was merged previously.

### DIFF
--- a/src/core/resources/ens/ensRegistration.ts
+++ b/src/core/resources/ens/ensRegistration.ts
@@ -1,0 +1,111 @@
+import { keccak256 } from '@ethersproject/keccak256';
+import { toUtf8Bytes } from '@ethersproject/strings';
+import { useQuery } from '@tanstack/react-query';
+import uts46 from 'idna-uts46-hx';
+
+import { ensClient } from '~/core/graphql';
+import {
+  QueryConfig,
+  QueryFunctionArgs,
+  QueryFunctionResult,
+  createQueryKey,
+} from '~/core/react-query';
+
+const ENS_DOMAIN = '.eth';
+
+function normalizeENS(name: string) {
+  try {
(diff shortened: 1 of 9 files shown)


greg-schrammel (approved):
nicee

src/core/resources/ens/ensRegistration.ts
```diff
@@ -0,0 +1,111 @@
...
+const fetchRegistration = async (ensName: string) => {
+  const response = await ensClient.getRegistration({
+    id: labelhash(ensName.replace(ENS_DOMAIN, '')),
+  });
+  const data = response.registration;
+
+  return {
+    registration: {
+      expiryDate: data?.expiryDate as string,
+      registrationDate: data?.registrationDate as string,
```
greg-schrammel, line 60: I think this type technically is `string | undefined`, is the `as string` needed? afraid someone assumes it always has a value in the future

estebanmino: i agree with this, we should have a fallback or nothing at all (?) if this fails

derHowie: Yep, this makes sense. I initially did this because these gql values are assigned as `any`. I have now changed this type to `string | undefined`.

estebanmino (approved):
🚀

PR #1177, opened 2023-12-06
Network Autocomplete

brunobar79:
Fixes BX-####
Figma link (if any):

Added an autocomplete for the network name field on the custom RPC form.

http://recordit.co/gvpvlcpFio


src/entries/popup/components/Autocomplete/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export interface AutocompleteProps {
+  data: AutocompleteData;
+  value: string;
+  onChange: (value: string) => void;
+  onSelect: (value: string) => void;
+  onBlur: () => void;
+  onFocus: () => void;
+  borderColor?: string;
```
greg-schrammel, line 25:
```suggestion
  borderColor?: BoxStyles['borderColor'];
```

src/entries/popup/components/Autocomplete/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+              </Command.List>
+            </Box>
+          )}
+        </Command>
+      </Box>
+    );
+  },
+);
+
+Autocomplete.displayName = 'Autocomplete';
```
greg-schrammel, line 155: jfyi another option to setting a displayName is using a named function in the forwardRef like
```ts
export const Autocomplete = forwardRef(function Autocomplete(props, ref) {
 ...
})
```

brunobar79: Good to know! That's probably cleaner than using display name. We should probably fix it across the entire codebase later on

src/entries/popup/components/Autocomplete/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+                width: '100%',
+                boxShadow: '2px 12px 12px rgba(0, 0, 0, 0.65)',
+                borderRadius: defaultBorderRadius,
+              }}
+            >
+              <Command.List>
+                {Object.keys(data).map((key) => {
+                  return (
+                    <Command.Group
+                      heading={key}
```
estebanmino, line 103: i see the keys of data object are also being used as string for heading here, any option to update it?

PR #1185, opened 2023-12-07
fix: duplicated supported chain + remove chain

estebanmino:
Fixes BX-####
Figma link (if any):

if you have a custom chain with an id that we already support, providers will blow up since we're using two chains with the same chainId as active to wagmi

so in this pr i'm filtering out any rnbw supported chain that is included in user's custom network

image

- fixed remove chain crash
-


greg-schrammel (approved):
makes sense 👍

PR #1197, opened 2023-12-12
Prevent Periodic NFT dups

derHowie:
In cases where we filtered out 50/50 polygon collections in a simple hash collections call, we would then proceed with the remaining logic and pass in an empty collection id array. This would sometimes cause duplicate NFTs to appear in the collection.

## What to test
This was fairly uncommon and didn't always happen so it's hard to illustrate a before and after. Please check that the logic looks alright and ensure you don't happen to run into the issue.


brunobar79 (approved):
🤝

greg-schrammel (approved):

PR #1199, opened 2023-12-12
tx error alerts

estebanmino:
Fixes BX-####
Figma link (if any):


src/entries/popup/handlers/wallet.ts
```diff
@@ -236,24 +228,22 @@ export const lock = async () => {
...
 };
 
 export const wipe = async () => {
   await walletAction('wipe', {});
-  await SessionStorage.set('userStatus', 'NEW');
-  return;
-};
-export const testSandbox = async () => {
-  return await walletAction('test_sandbox', {});
+  return walletAction('test_sandbox', {});
```
brunobar79, line 240: I think this doesn't belong here.

brunobar79 (approved):

src/entries/popup/handlers/wallet.ts
estebanmino: uh oh

PR #1216, opened 2023-12-15
wagmi patch

brunobar79:
Fixes BX-####
Figma link (if any):

Removing some stuff that we don't use and while it's being removed by Webpack's treeshaking
I want to avoid any confusions at the lockfile level


derHowie (commented):
lgtm ✅

PR #1231, opened 2023-12-18
send customrpc checks

estebanmino:
Fixes BX-1194
Figma link (if any):

- adding send checks for custom rpcs
- fetching native assets for tx fee, there was a bug where users with a token balance but no native asset balance would see "undefined" as tx fee when trying to send a token since we were filtering out native assets with balance 0

image


estebanmino:
after talking with @brunobar79 i added the `sending on network` ui but it won't show an explainer

image

brunobar79 (approved):
🤝 🤝 🤝

derHowie (approved):

PR #1232, opened 2023-12-18
fix: reorg custom rpcs to rainbow chains

estebanmino:
Fixes BX-####
Figma link (if any):

the problem causing this bug was that the rainbow supported networks weren't being included in the custom chain object, so if you added a rpc with chain id 1, the code was picking the state and saying "there's no active rpc url for this network", so it was switched to active automatically

so now we're adding the rnbw supported networks to that state too

Edits from 19 dec:

having an state called `customRPCs` and `customRPCAssets` doesn't make any sense since we're also storing information from the networks that rainbow supports by default.

Now we have `rainbowChains` and `rainbowChainAssets` which are more suitable for the model we currently have on the ui.

a `RainbowChain` has multiple `chains` that correspond to the default rainbow chain + any custom rpc the user wants to add, and the active rpc url which is being used. This opens the door to have more specific networks configuration like `name`

a `RainbowChainAssets` slice contains assets to an specific `RainbowChain` by chainId, it doesn't depend on any custom rpc chain that the user added


src/core/state/rainbowChains/index.ts
```diff
@@ -0,0 +1,142 @@
...
+} from '~/core/types/chains';
+
+import { createStore } from '../internal/createStore';
+
+export interface RainbowChain {
+  activeRpcUrl: string;
+  chains: Chain[];
+}
+
+export interface rainbowChainstate {
```
brunobar79, line 18: rainbowChainState ?

brunobar79 (approved):
🐐 🐐 🐐 🐐 🐐 🐐 🐐 🐐 🐐 🐐 🐐

src/core/state/rainbowChains/index.ts
estebanmino: uh

changing to `RainbowChainState`

PR #1238, opened 2023-12-19
[BX-1193] Custom Network State Restoration

derHowie:
Added State Restoration to Add Network, Add RPC, and Add Token forms.

Add Network/RPC PoW: https://recordit.co/Neo37guoYh
Add Token PoW: https://recordit.co/aMxMRobeJl

## What to test
As with other pieces of state restoration data, these fields will not persist any draft data when the window is full screen, so make sure to test using the popup only.

Make sure that adding networks, rpcs, and tokens still behaves correctly. Ensure that drafts are being saved when navigation away from each of the forms, but also make sure that the draft data is deleted once the add action is performed.


src/entries/popup/hooks/useRestoreNavigation.ts
```diff
@@ -74,7 +74,11 @@ const RESTORE_NAV_MAP: Record<string, string[]> = {
   [ROUTES.SETTINGS__NETWORKS]: [ROUTES.SETTINGS, ROUTES.SETTINGS__NETWORKS],
   [ROUTES.SETTINGS__NETWORKS__CUSTOM_RPC]: [
     ROUTES.SETTINGS,
-    ROUTES.SETTINGS__NETWORKS__CUSTOM_RPC,
+    ROUTES.SETTINGS__NETWORKS,
```
derHowie, line 77: State restoration doesn't currently support dynamic routes, so this line was actually causing a little bug. This would take users back to the add network form instead of the add rpc form, because the chain id location state does not exist after restoration.

Might be worth going back and supporting this later, but for now I figured best to just drop the user back on the network settings page.

src/entries/popup/hooks/useRestoreNavigation.ts
```diff
@@ -74,7 +74,11 @@ const RESTORE_NAV_MAP: Record<string, string[]> = {
...
+  ],
+  [ROUTES.SETTINGS__NETWORKS__CUSTOM_RPC__DETAILS]: [
+    ROUTES.SETTINGS,
+    ROUTES.SETTINGS__NETWORKS,
```
derHowie, line 81: Noticed that on this page I was getting dropped all the way back to home, so I'm at least getting the user a little bit closer to where they left off.

src/entries/popup/pages/settings/customChain/index.tsx
```diff
@@ -508,12 +526,16 @@ export function SettingsCustomChain() {
...
-          active: true,
-        }));
+        setCustomRPC((prev) => {
+          const newState = {
+            ...prev,
+            ...network.value,
+            name: networkName,
+            active: true,
+          };
+          saveCustomNetworkDraft(draftKey, newState);
```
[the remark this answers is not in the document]
derHowie: ok, moved all of it to useEffect

src/entries/popup/hooks/useRestoreNavigation.ts
brunobar79: what if we change the url to pass the chain via query string? That should be a very small change and would fix this issue completely.

estebanmino (approved):

src/entries/popup/hooks/useRestoreNavigation.ts
derHowie: https://recordit.co/tamNg9gQ4I

Alright, went with persisting location.state but this worked in fixing the paths for both routes

src/entries/popup/hooks/useRestoreNavigation.ts
brunobar79: Just for the record, this has been fixed in future commits

brunobar79 (approved):
💯 💯 💯

PR #1239, opened 2023-12-19
[BX-1198][BX-1199] NFTs Disappearing after Adding Custom RPC

derHowie:
I hardcoded a list of network names that simplehash supports to prevent 400s when attempting to use custom network names in the nfts calls.

PoW: https://recordit.co/XITrUciUq5

## What to test
Add a nonsense custom rpc and make sure nfts still show up


src/core/resources/nfts/nfts.ts
```diff
@@ -51,9 +54,12 @@ async function nftsQueryFunction({
   if (process.env.IS_TESTING === 'true') {
     return NFTS_TEST_DATA;
   }
-  const chains = getBackendSupportedChains({ testnetMode: false }).map(
-    ({ name }) => name as ChainName,
-  );
+  const simpleHashSupportedChains = getSimpleHashSupportedChainNames();
+  const chains = getBackendSupportedChains({ testnetMode: false })
+    .map(({ name }) => name as ChainName)
+    .filter((chainName) =>
```
estebanmino, line 60: could we hardcode these?

i think the problem is coming from `getBackendSupportedChains` since is using `getNetwork` which takes all of the networks from wagmi provider

i feel like this error could be also happening when getting assets and txs from backend

estebanmino: maybe using SUPPORTED_MAINNET_CHAINS is enough

estebanmino: and SUPPORTED_TESTNET_CHAIN_IDS for testnets

derHowie: Aren't we preventing users from discovering NFTs elsewhere if we just use SUPPORTED_MAINNET_CHAINS and SUPPORTED_TESTNET_CHAINS rather than attempting to use the additional networks they've added? Or do we not really care about that?

derHowie: Yes, `getNetwork` grabbing the chain names via wagmi was the problem for two reasons. Some custom chains aren't supported by simplehash.. also when a user adds a supported chain with a different name/nickname that also fucks up the nft call.

brunobar79: Just for context: RE: Esteban's comment - `i feel like this error could be also happening when getting assets and txs from backend` -  This was true but I specifically asked backend to ignore chains they don't support and still return a valid response. They did that for assets (and I think they also did it for txs)

estebanmino: i believe we do we not really care about that, but wdyt? if you feel confident that we can support nfts of custom networks i['m down

brunobar79: I chatted with Chris and for now we're gonna go with only the supported ones like @estebanmino  suggested earlier.

estebanmino (approved):

brunobar79 (approved):
🤝🤝🤝

PR #1240, opened 2023-12-19
Fix blockexplorer for custom chains

brunobar79:
Fixes BX-1196
Figma link (if any):

We were not saving the block explorer url for custom chains

https://recordit.co/kJkYzkffRt


estebanmino (approved):

src/entries/popup/pages/settings/customChain/index.tsx
```diff
@@ -457,6 +457,14 @@ export function SettingsCustomChain() {
           name: symbol,
         },
         rpcUrls: { default: { http: [rpcUrl] }, public: { http: [rpcUrl] } },
+        blockExplorers: {
+          default: {
+            name: customRPC.explorerUrl
+              ? getDappHostname(customRPC.explorerUrl)
+              : '',
+            url: customRPC.explorerUrl || '',
```
greg-schrammel, line 465: if there is no explorerUrl should we do `''`? can that show something like `view on  ` like "view on >blank<" anywhere?

greg-schrammel (approved):

src/entries/popup/pages/settings/customChain/index.tsx
brunobar79: We already have checks on the UI side for this. I think wagmi types were complaining about undefined

PR #1241, opened 2023-12-19
Slow down URL validation when adding a network or RPC

greg-schrammel:
Fixes BX-####
Figma link (if any):

ups a bit the debounce delay and adds `https://` to the url if needed


estebanmino (approved):

src/entries/popup/pages/settings/customChain/index.tsx
```diff
@@ -277,6 +277,8 @@ const KNOWN_NETWORKS = {
   ],
 };
 
+const parseURL = (url: string | undefined) =>
+  url && (url.startsWith('http') ? url : `https://${url}`);
```
brunobar79, line 281: If a user enters `localhost:8545` you'll autocomplete it to https://localhost:8545  which won't work so we have 2 options:

- Drop this and not add the prefix
- Add an exception for hosts `localhost` & `127.0.0.1` where you only add `http://`

brunobar79 (changes requested):

src/entries/popup/pages/settings/customChain/index.tsx
brunobar79: I'm removing it for now so I can merge it

PR #1250, opened 2023-12-20
fix: chain names

estebanmino:
Fixes BX-1205
Figma link (if any):

fixing BX-1205 again and including all of the places we use chain name


brunobar79 (approved):
🤝🤝🤝

PR #1257, opened 2023-12-21
hide chain id

brunobar79:
Fixes BX-####
Figma link (if any):

We don't need to expose this field at all.

https://recordit.co/18tFU4Ba7p


estebanmino (approved):
lfggg

PR #1262, opened 2023-12-27
fix for custom network assets

brunobar79:
Fixes BX-####
Figma link (if any):

Backend was finding assets for custom networks but those were incomplete (missing chain id). 
That would throw an error an exit the middle of the list showing an incomplete asset list.

I've fixed the missing props and added some safety checks to prevent bad requests.

Screen Shot 2023-12-27 at 3 33 47 PM

## What to test

You were able to repro the issue by following these steps:

on a fresh install

watch vitalik.eth
add + enable avalanche from suggested networks list
wallet data stops showing


src/core/utils/assets.ts
```diff
@@ -83,7 +84,11 @@ export function parseAsset({
 }): ParsedAsset {
   const address = asset.asset_code;
   const chainName = asset.network ?? ChainName.mainnet;
-  const chainId = chainIdFromChainName(chainName);
+  const chainId =
+    chainIdFromChainName(chainName) ||
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    (Number(Object.keys(asset.networks)?.[0]) as ChainId);
```
brunobar79, line 91: Backend isn't returning a chainId for these assets but the chain id was available under the networks dict.

src/core/utils/assets.ts
```diff
@@ -113,7 +118,7 @@ export function parseAsset({
     symbol: asset.symbol,
     type: asset.type,
     decimals: asset.decimals,
-    icon_url: asset.icon_url,
+    icon_url: asset.icon_url || getCustomChainIconUrl(chainId, address),
```
brunobar79, line 121: also those assets discovered by the backend don't have an icon but we can try to get it anyways

derHowie (approved):
🧹

src/entries/popup/hooks/send/useSendAsset.ts
```diff
@@ -65,7 +66,15 @@ export const useSendAsset = () => {
   );
 
   const allAssets = useMemo(
-    () => [...assets, ...customNetworkAssets],
+    () =>
+      sortedUniqBy(
+        [...assets, ...customNetworkAssets].sort(
+          (a: ParsedUserAsset, b: ParsedUserAsset) =>
+            parseFloat(b?.native?.balance?.amount) -
```
greg-schrammel, line 73: `parseFloat(undefined)` is `NaN`, NaN is treated as 0 in `sort`, not sure if it matters just to be sure it's expected, I can think it may happen of a token where balance.amount is undefined somehow is sorted above another because of this

greg-schrammel (approved):
👍

src/entries/popup/hooks/send/useSendAsset.ts
brunobar79: Fair point but even with NaN it's the expected behavior (undefined treated as 0 and pushed to the end)

PR #1275, opened 2024-01-02
fix: BX-1220

estebanmino:
Fixes BX-1220
Figma link (if any):

for empty wallets with no assets we're not fetching the native assets so this bug was happening

with this pr i'm introducing mock assets to have the native assets with prices coming from backend but balance being 0 so the user will see the tx fee correctly

https://www.loom.com/share/85086669138243bdbcbcb1ba0415e8a3

### DIFF
--- a/src/entries/popup/hooks/useNativeAsset.ts
+++ b/src/entries/popup/hooks/useNativeAsset.ts
@@ -1,15 +1,44 @@
-/* eslint-disable no-nested-ternary */
 import { AddressZero } from '@ethersproject/constants';
 import { Address, useNetwork } from 'wagmi';
 
+import { NATIVE_ASSETS_MAP_PER_CHAIN } from '~/core/references';
 import { useUserTestnetNativeAsset } from '~/core/resources/assets/userTestnetNativeAsset';
 import { useCurrentAddressStore, useCurrentCurrencyStore } from '~/core/state';
+import { ParsedUserAsset } from '~/core/types/assets';
 import { ChainId } from '~/core/types/chains';
-import { isCustomChain } from '~/core/utils/chains';
+import { chainNameFromChainId, isCustomChain } from '~/core/utils/chains';
 
 import { useCustomNetworkAsset } from './useCustomNetworkAsset';
 import { getNetworkNativeAssetUniqueId } from './useNativeAssetForNetwork';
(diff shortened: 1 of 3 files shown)


src/entries/popup/hooks/useNativeAsset.ts
```diff
@@ -1,23 +1,52 @@
...
+}: {
+  chainId: ChainId;
+}): ParsedUserAsset | undefined | null => {
+  const nativeAssets = useNativeAssets();
+  const { chains } = useNetwork();
+  if (!nativeAssets) return null;
+  return chains
+    .map((chain) => {
+      const assetKey = `${NATIVE_ASSETS_MAP_PER_CHAIN[chain.id]}_${
+        ChainId.mainnet
```
greg-schrammel, line 27: is mainnet right?

src/entries/popup/hooks/useNativeAsset.ts
```diff
@@ -1,23 +1,52 @@
...
+        ...nativeAsset,
+        chainId: chain.id,
+        chainName: chainNameFromChainId(chain.id),
+        native: {
+          balance: { amount: '0', display: `0 ${nativeAsset?.symbol}` },
+        },
+        balance: { amount: '0', display: `0 ${nativeAsset?.symbol}` },
+      };
+    })
+    .find((c) => c.chainId === chainId);
```
greg-schrammel, line 40: hm if we're filtering for the chain in the end is building the whole map even necessary?

src/entries/popup/hooks/useNativeAsset.ts
estebanmino: yes this is coming from here https://github.com/rainbow-me/browser-extension/blob/3d5e28eed26cc6e8d6c6ee4ec1b7e1e8bcd362eb/src/entries/popup/hooks/useNativeAssets.ts#L35C21-L35C21

src/entries/popup/hooks/useNativeAsset.ts
estebanmino: i'll do this at the beginning

brunobar79 (approved):
Good idea 🤝

greg-schrammel (approved):
👑

PR #1283, opened 2024-01-04
[BX-904] Track CommandK Commands

derHowie:
Tracking command id, label, and name upon execution

## What to test
TBD


greg-schrammel (approved):

src/entries/popup/components/CommandK/utils.ts
```diff
@@ -240,6 +242,12 @@ export function useCommandExecution(
 
   const handleExecuteCommand = React.useCallback(
     (command: SearchItem | null, e?: KeyboardEvent) => {
+      analytics.track(event.commandKActionExecuted, {
+        id: command?.id,
+        label: command?.actionLabel,
+        name: command?.name,
```
DanielSinclair, line 248: @derHowie Thought of an privacy edge case with the wallet search items. In that case, I think we'd log the wallet name/ENS/address in `label` and `name`. For the wallet, token, nft search item types, could we just make the `name` something like `View Wallet` and leave the `label` undefined

derHowie: Alright. I updated this to excluded name and id from 'View Wallet', 'Open [NFT]', and 'Open [Token]'. Lmk if the diff covers what you had in mind.

DanielSinclair (approved):

DanielSinclair:
@derHowie Added an approval! lgtm

PR #1286, opened 2024-01-08
fix missing activity on user wallet

greg-schrammel:
Fixes BX-####
Figma link (if any):

pow

https://github.com/rainbow-me/browser-extension/assets/6232729/9e47d66b-4e23-4f25-b37f-78b04fe08739


src/core/utils/assets.ts
```diff
@@ -100,18 +100,16 @@ export function parseAsset({
 }): ParsedAsset {
   const address = asset.asset_code;
   const chainName = asset.network ?? ChainName.mainnet;
+  const networks = 'networks' in asset ? asset.networks || {} : {};
   const chainId =
-    chainIdFromChainName(chainName) ||
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    (Number(Object.keys(asset.networks)?.[0]) as ChainId);
+    chainIdFromChainName(chainName) || Number(Object.keys(networks)[0]);
```
greg-schrammel, line 105: some assets don't return a `networks` this was erroring in some custom networks and making so that nothing was shown in the activity tab, Im talking with Benjamin to find a better solution

src/core/utils/transactions.ts
```diff
@@ -153,10 +153,10 @@ export function parseTransaction({
     ...change,
     asset: parseUserAsset({
       asset: change.asset,
-      balance: change.value?.toString() || '0',
+      balance: change.quantity || '0',
```
greg-schrammel, line 156: `value` was deprecated and `quantity` should be used now

src/core/utils/assets.ts
estebanmino: we got new updates from backend  now, would it fix this?

derHowie (approved):

src/core/utils/assets.ts
greg-schrammel: yes updated to use the new `chain_id` field from the api response

estebanmino (approved):

DanielSinclair:
@SocketSecurity ignore-all

PR #1291, opened 2024-01-08
[BX-1244] Add a link to Transactions > use Flashbots > learn more

BrodyHughes:
noticed we just have a spot that says 'learn more' and didn't link anywhere. this should fix that.

Screenshot 2024-01-08 at 2 37 43 PM


src/entries/popup/components/Menu/MenuItem.tsx
```diff
@@ -116,12 +118,14 @@ interface DescriptionProps {
   color?: TextStyles['color'];
   text: string;
   weight?: TextStyles['fontWeight'];
+  descriptionLink?: () => void;
```
estebanmino, line 121: IMO this name should be `onClickLink` or something like that

as `descriptionLink` i understood is a string containing the link, not the action

BrodyHughes: That makes sense to me. will change.

DanielSinclair (approved):

static/json/languages/en_US.json
```diff
@@ -231,7 +231,7 @@
       "use_flashbots": "Use Flashbots",
       "clear_transactions_and_nonces": "Clear Pending Transactions",
       "clear_transactions_and_nonces_success": "Pending transactions cleared",
-      "flashbots_description": "Flashbots protects your transactions and reduces the likelihood of transaction failures. Learn more."
+      "flashbots_description": "Flashbots protects your transactions and reduces the likelihood of transaction failures. "
```
DanielSinclair, line 234: The trailing space may get dropped when translating, so may need to handle the padding differently

estebanmino (approved):

PR #1296, opened 2024-01-10
fix pending txs

estebanmino:
Fixes BX-1087
Fixes BX-1224
Fixes BX-1233
Fixes BX-1048
Fixes BX-1230
Fixes BX-1271

Figma link (if any):

- refactored the way we use pendingTransaction state slice

  for some reason we had the `getPendingTransaction` method inside the state parsing `NewTransation` to `RainbowTransaction` when getting the txs so I refactored the code to only store `RainbowTransaction` so we don't have to be aware of that parsing step.
  I also removed the need of `getPendingTransaction` changing the state to be able to get the pending txs as a hook as `pendingTransactions[address]` so the ui updates correctly to the `Clear pending txs` button in settings

- `getNextNonce` will depend on provider request first to get the next nonce, to solve the flashbots dropped issue where the user is not able to send any other tx

- refactored `watchPendingTx` method, now is divided in custom chain, flashbots and supported chain id txs with different methods. `getTransactionReceipt` resolves with null after one second if the provider is not able to get the receipt, this was part of the flashbot bug

    - sent a flashbots tx with nonce 10
    - sent a tx with nonce 11
    - sent a tx with nonce 12

    we have 3 pending txs, if the tx with nonce 10 gets dropped, we're awaiting forever for `getTransactionReceipt` to resolve and therefore we never update the pending txs state https://linear.app/rainbow/issue/APP-951/clear-pending-transactions-button-doesnt-work#comment-4511f2b4

EDIT Jan 17

- Added flashbots flag for send / approval txs when flashbots is enabled, this was missing so we weren't checking status directly through flashbots provider

EDIT Jan 19

adding a fix for this as well https://linear.app/rainbow/issue/BX-1271/issue-dark-screen-overlay-happens-when-attempting-to-cancelspeed-up
thw activity view wasn't being closed after interacting with the speedup cancel sheet

image

sent multiple txs on mainnet, mainnet with flashbots, op, custom network to check this working

flashbots error fix https://www.loom.com/share/e86cc74c121445b6bef7a803af9e708d


src/core/utils/transactions.ts
```diff
@@ -310,16 +331,13 @@ export async function getNextNonce({
...
-  const localNonce = localNonceData?.currentNonce || 0;
   const provider = getProvider({ chainId });
   const txCountIncludingPending = await provider.getTransactionCount(
     address,
     'pending',
   );
-  if (!localNonce && !txCountIncludingPending) return 0;
-  const ret = Math.max(localNonce + 1, txCountIncludingPending);
+  if (!txCountIncludingPending) return 0;
+  const ret = txCountIncludingPending;
```
estebanmino, line 340: depending on provider for nonces

src/entries/popup/hooks/useInfiniteTransactionList.ts
```diff
@@ -31,9 +31,13 @@ export const useInfiniteTransactionList = ({
 }: UseInfiniteTransactionListParams) => {
   const { currentAddress: address } = useCurrentAddressStore();
   const { currentCurrency: currency } = useCurrentCurrencyStore();
-  const { getPendingTransactions } = usePendingTransactionsStore();
+  const { pendingTransactions: storePendingTransactions } =
+    usePendingTransactionsStore();
   const [manuallyRefetching, setManuallyRefetching] = useState(false);
-  const pendingTransactions = getPendingTransactions({ address });
```
estebanmino, line 36: calling `getPendingTransactions` is not allowing `pendingTransactions` to update inside the component since is not listening to state

src/entries/popup/hooks/useWatchPendingTransactions.ts
```diff
@@ -22,149 +21,182 @@ import {
...
+    },
+    [address, currentCurrency],
+  );
+
+  const processPendingTransaction = useCallback(
+    async (tx: RainbowTransaction) => {
+      let updatedTransaction: RainbowTransaction = { ...tx };
+      try {
+        const chainId = tx?.chainId;
+        if (chainId && tx.hash && address) {
```
estebanmino, line 113: processing txs depending on tx type

src/entries/popup/hooks/useWatchPendingTransactions.ts
```diff
@@ -22,149 +21,182 @@ import {
...
+
+      chainIds.map(async (chainId) => {
+        const provider = getProvider({ chainId });
+        const providerTransactionCount = await provider.getTransactionCount(
+          address,
+          'latest',
+        );
+        const currentProviderNonce = providerTransactionCount - 1;
+        const currentNonceForChainId = highestNoncePerChainId.get(chainId);
+        setNonce({
```
estebanmino, line 179: we were doing this per tx before, now we do it per chainId in pendingTxs

derHowie (approved):
🙏 

Code looks good, didn't see any pending tx hang. I did notice that there is a small gap between the pending tx disappearing and the confirmed transaction showing up. Example: https://recordit.co/35PEcKPU0i (they did show up a little less than a minute later 👍)

estebanmino:
thanks for the review @derHowie ! taking a look

estebanmino:
@derHowie just fixed the bug you mentioned before https://www.loom.com/share/650ea924b422470b9d646ca0185a3cb4

brunobar79 (approved):
💯 💯 💯

BrodyHughes (approved):
lgtm

PR #1298, opened 2024-01-12
block add chain duplicates

estebanmino:
Fixes BX-1231
Figma link (if any):

implementing ticket

https://www.loom.com/share/e2027f611c6546ea99f63dd70146a9b9

EDIT

@brunobar79 pointed out the case where you are adding a rpc that you want like flashbots, in which case we shouldn't block and set as active instead

this pr is not:

- adding new custom rpc and setting it as active if the rpc url is not present
- setting new custom rpc as active if the rpc url is present but is not currently active

https://www.loom.com/share/df682b11e3374bbca9fc4616c6b55440


derHowie (approved):
👌 🔥

brunobar79 (approved):
👍 👍 👍

PR #1306, opened 2024-01-17
Custom network e2e

BrodyHughes:
Fixes BX-1077
- Auto complete adding of a new network
- Adding a custom network
- Adding a Custom Testnet (the testnet checkboxes)
- Adding an RPC
- Adding a custom token

I can add sending on a custom network and token discovery in the future. The above flows cover the basics and keep the scope small for now. it's currently in the `send` folder bc i plan on adding sends soon. Can move it for now if y'all want.


DanielSinclair:
@BrodyHughes Looks good so far! We might want to move to a new `customNetworks` serial folder because I think the scope of this might expand as we test more things like Custom Network NFT discovery, etc./

DanielSinclair (approved):

estebanmino (approved):

PR #1311, opened 2024-01-22
fix: pending tx details

estebanmino:
Fixes BX-1279
Figma link (if any):

the code is expecting that BE returns valid tx data for pending txs which is not correct

this pr checks if the BE is sending `status` `pending` to go and get the tx from provider

https://www.loom.com/share/f6ead3ce5f634cc8b666c116e721aa04


brunobar79 (approved):
💯

derHowie (approved):

PR #1324, opened 2024-01-29
fix: block non supported networks for some requests

estebanmino:
Fixes BX-####
Figma link (if any):

https://rainbowhaus.slack.com/archives/C044WE28BTK/p1706550952588949

blocking these requests when chain id is not supported


derHowie (approved):

src/core/utils/chains.ts
```diff
@@ -83,6 +83,10 @@ export const getSupportedChainsWithHardhat = () => {
   );
 };
 
+export const isDefaultSupportedChain = ({ chainId }: { chainId: ChainId }) => {
+  return SUPPORTED_CHAINS.map((chain) => chain.id).includes(chainId);
+};
```
greg-schrammel, line 88: we got a `SUPPORTED_CHAIN_IDS` already why not use that?

greg-schrammel (approved):

src/core/utils/chains.ts
estebanmino: i'm updatng now

brunobar79 (approved):

PR #1329, opened 2024-01-30
Updated script to point to a specific version of ff dev edition

BrodyHughes:
- Updated script to point to a specific version of ff dev edition.

- Changes to the `SettingsFlows` test and `ShortcutSettingsFlows` were due to the default toggle behavior being changed on Firefox while these tests were broken.


brunobar79 (approved):
🔥🔥🔥

DanielSinclair (approved):

PR #1297, opened 2024-01-12
fix: BX-1256

estebanmino:
Fixes BX-####
Figma link (if any):

## What changed (plus any additional context for devs)

transforming bn values to strings

PR #1330, opened 2024-01-30
[BX-1281] Send hex strings instead of BNs through provider messenger

derHowie:
Current working theory is that frontends that don't load ethers/bignumber have an issue accepting big number objects that have been serialized by the messenger bridge.

#1297 addressed some of these cases but there were still some big number objects preventing the transaction flow from working correctly on blur (and presumably other dapps).

Blur was blowing up after using the transaction payload we were sending back. The transaction was succeeding but the dapp UI would hang indefinitely.

Successful TX on Blur.io without hanging: https://recordit.co/oBwmvKwPeG

## What to test
Let me know if there are other considerations I should be making and whether these changes have broader implications.


brunobar79 (approved):
💯

DanielSinclair (approved):

PR #1338, opened 2024-02-07
rename on watch wallet

greg-schrammel:
Fixes BX-####
Figma link (if any):

https://github.com/rainbow-me/browser-extension/assets/6232729/3b69196f-7940-414b-a055-43675b8a9726


estebanmino (approved):
lgtm

DanielSinclair:
@greg-schrammel Noticed that when you paste an address that resolves to an ENS, it still requires a name; I think we would just want to add that logic so you only need to name a wallet when an address does not resolve

DanielSinclair (changes requested):
@greg-schrammel Noticed that when you paste an address that resolves to an ENS, it still requires a name; I think we would just want to add that logic so you only need to name a wallet when an address does not resolve

BrodyHughes (approved):

DanielSinclair (approved):

PR #1343, opened 2024-02-10
fix: Don't route the rpcUrl through proxy if it's on a local IP range

kryptopaul:
Fixes BX-####
Figma link (if any): -

Made `src\core\providers\proxy.ts` not proxy the traffic if the provided `rpcUrl` is running within a local IP range.

Context:

I was trying to add a RPC running on my local network (but not on my `localhost`) and Rainbow would throw the `rpc_not_responding` error. Upon looking into the code, I was able to see how that error would be caused by `useChainMetadata()` in `src\entries\popup\pages\settings\customChain\index.tsx`.

A few function calls deep, I've reached `getChainMetadataRPCUrl()` (L323) in `src\core\utils\chains.ts`. I've noticed that in L329, the `rpcUrl` is passed to `proxyRpcEndpoint()` before being further used by `JsonRpcProvider` from Ethers.

Within `proxyRpcEndpoint()` it seems that the code checks whether the RPC is running on `localhost` or `127.0.0.1`. If that is the case, it leaves the URL as it is. If it is something different, it appears to proxy the traffic through your servers (?) before sending the requests to the actual RPC.

This will not work if the provided URL is on user's local network (but not on `localhost`). If the IP starts with 10, 192 or 172, it should not be routed through your proxy.

image

## What to test

Try to add a custom RPC which runs on your local network, but not on the same device.

### DIFF
--- a/src/core/providers/proxy.ts
+++ b/src/core/providers/proxy.ts
@@ -3,7 +3,10 @@ import { ChainId } from '../types/chains';
 export const proxyRpcEndpoint = (endpoint: string, chainId: ChainId) => {
   if (
     endpoint !== 'http://127.0.0.1:8545' &&
-    endpoint !== 'http://localhost:8545'
+    endpoint !== 'http://localhost:8545' &&
+    !endpoint.includes('http://10.') &&
+    !endpoint.includes('http://192.') &&
+    !endpoint.includes('http://172.')
   ) {
     return `${process.env.RPC_PROXY_BASE_URL}/${chainId}/${
       process.env.RPC_PROXY_API_KEY


src/core/providers/proxy.ts
```diff
@@ -3,7 +3,10 @@ import { ChainId } from '../types/chains';
 export const proxyRpcEndpoint = (endpoint: string, chainId: ChainId) => {
   if (
     endpoint !== 'http://127.0.0.1:8545' &&
-    endpoint !== 'http://localhost:8545'
+    endpoint !== 'http://localhost:8545' &&
+    !endpoint.includes('http://10.') &&
+    !endpoint.includes('http://192.') &&
+    !endpoint.includes('http://172.')
```
greg-schrammel, line 9: IANA private ips range is 192.168, so 192.16**9**.x.x can be public
and 172. ranges from 172.16.x.x to 172.31.x.x

maybe something like this
```suggestion
    !endpoint.includes('http://192.168') &&
    !endpoint.match(/http:\/\/172.(1[6-9]|2[0-9]|3[0-1])./)
```

kryptopaul: Right. This seems okay.

DanielSinclair:
@brunobar79 For review and external PR admin merge

brunobar79 (approved):
LGTM! Thanks for doing this.

brunobar79 (approved):
LGTM! Thanks for doing this.

PR #1368, opened 2024-02-28
[BX-1294] Fix missing wallets in Send flow

BrodyHughes:
Fixes BX-1294

- was previously just checking the order of wallets in `useWalletOrderStore()` and returning those. this was only ever updated when the order was changed in the walletSwitcher.
- I updated the send flow to check the existing list of wallets in `getAccounts()` and then compare that to the order of the wallets in `useWalletOrderStore()`
- it then updates the `useWalletOrderStore()` with all the wallets in the correct order and also updated the `useWalletOrderStore()`

## What to test
- see if send suggestions includes all the wallets in your bx aside from the current wallet
- make sure code makes sense

<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/1d09df63-322f-4255-8be3-dbe94c520328" >


src/entries/popup/pages/send/ToAddressInput.tsx
```diff
@@ -161,15 +162,45 @@ const DropdownWalletsList = ({
...
   selectWalletAndCloseDropdown: (address: Address) => void;
 }) => {
-  const { walletOrder } = useWalletOrderStore();
+  const [newWalletOrder, setNewWalletOrder] = useState<`0x${string}`[]>([]);
+  const { walletOrder, saveWalletOrder } = useWalletOrderStore();
+
+  // find the items in getAccounts then compare to the current
+  // walletOrder, then return all items in getAccounts in order
+
+  useEffect(() => {
```
greg-schrammel, line 171: have you tried the `useAccounts` hook?

DanielSinclair (approved):
Functionality looks good to me; didn't review code

src/entries/popup/pages/send/ToAddressInput.tsx
```diff
@@ -161,15 +162,45 @@ const DropdownWalletsList = ({
...
-    [wallets, walletOrder],
+    () => sortWallets(newWalletOrder, wallets),
+    [newWalletOrder, wallets],
   );
   const sortedWatchedWallets = useMemo(
-    () => sortWallets(walletOrder, watchedWallets),
-    [watchedWallets, walletOrder],
+    () => sortWallets(newWalletOrder, watchedWallets),
+    [newWalletOrder, watchedWallets],
   );
```
greg-schrammel, line 202: after playing with it a bit I'm pretty sure the props are in the right state already, we just need to remove all sorting from inside this component
like 
Screenshot 2024-03-07 at 15 46 33

BrodyHughes: changed lol

greg-schrammel (approved):

PR #1370, opened 2024-02-29
feat: pin / unpin tokens

magiziz:
Fixes BX-1328
Figma link (if any):
- [Context menu design](https://www.figma.com/file/hzfwKOddfRmeVO4KUsG9wM/branch/aLkgDopprmzkQRc7M796PN/Rainbow-BX?type=design&node-id=16098-14055&mode=design&t=2dSNDGuOQ3lPuU5M-0)
- [Highlighter indicator design](https://www.figma.com/file/hzfwKOddfRmeVO4KUsG9wM/branch/aLkgDopprmzkQRc7M796PN/Rainbow-BX?type=design&node-id=16099-15553&mode=design&t=pCrl8f03u4whihbA-0)

https://github.com/rainbow-me/browser-extension/assets/53529533/cc5d837c-78d1-49d1-98de-61585579b401

## What to test
- Go to tokens tab on homepage
- Press the token row and make sure it pins at first and unpins if you press again
- When token is pinned it gets on top of the tokens list
- When right clicking you'll also see pin / unpin token option. Should work exactly the same, but without pressing.


src/entries/popup/pages/home/Tokens.tsx
```diff
@@ -123,22 +133,16 @@ export function Tokens() {
...
-            parseFloat(a?.native?.balance?.amount),
-        ),
-        'uniqueId',
-      ),
+  const combinedAssets = useMemo(
+    () => [...assets, ...customNetworkAssets],
     [assets, customNetworkAssets],
   );
 
+  const filteredAssets = useFilteredPinnedAssets(combinedAssets);
```
KosmosKey, line 141: Pinned tokens should be on top and the rest of tokens should appear after pinned tokens. Just a small filter function.

src/entries/popup/pages/home/Tokens.tsx
```diff
@@ -203,10 +207,12 @@ export function Tokens() {
...
           {assetsRowVirtualizer.getVirtualItems().map((virtualItem) => {
             const { key, size, start, index } = virtualItem;
-            const token = allAssets[index];
+            const token = filteredAssets[index];
+            const pinned = uniqueIds.some((id) => id === token.uniqueId);
+
             return (
               <Box
-                key={key}
+                key={`${token.uniqueId}-${key}`}
```
KosmosKey, line 215: When pressed the react state will reset because key changes. This is useful so that we're no longer in a "pressed" active state.

estebanmino (approved):
lgtm!

DanielSinclair:
A few tweaks!

- Can we bump the border to outside of the rounded cell? The appear animation should feel a bit better that way 
> Screenshot 2024-03-06 at 12 45 05 AM

- I noticed the pinned items get sorted; can we shift this to first-in-first-out pinning so that the order is persistent?

- Can we ignore click-drags (like clicking on cell and then dragging cursor to menu bar)? Noticed you can accidentally trigger the pinning when clicking and dragging; looks like the timing is a bit faster than hold-press

BrodyHughes (approved):
LGTM! Works well. Only thought is: do we want to limit token names to one line in the context menu using a `TextOverflow`?
Screenshot 2024-03-06 at 2 32 22 PM
Screenshot 2024-03-06 at 2 31 47 PM

KosmosKey:
@BrodyHughes Updated 🙏

image

src/entries/popup/pages/home/Tokens.tsx
```diff
@@ -123,22 +144,16 @@ export function Tokens() {
...
-    () =>
-      uniqBy(
-        [...assets, ...customNetworkAssets].sort(
-          (a: ParsedUserAsset, b: ParsedUserAsset) =>
-            parseFloat(b?.native?.balance?.amount) -
-            parseFloat(a?.native?.balance?.amount),
-        ),
-        'uniqueId',
-      ),
+  const combinedAssets = useMemo(
```
DanielSinclair, line 147: @KosmosKey Looks like there is a merge conflict with #1383 here and that new logic should be kept. I think we also typically perform filters in-line instead of a separate hook, so I'd suggest we refactor to shift to that. Maybe a `usePinnedAssets` hook would be broken out, and then the combined filters would be inline here

DanielSinclair (changes requested):
Added above comment about merge conflict and filter hooks

DanielSinclair:
@SocketSecurity ignore-all

DanielSinclair (approved):

PR #1372, opened 2024-02-29
Unpair HW wallets

BrodyHughes:
Fixes BX-1253
Figma link (if any): https://www.figma.com/file/hzfwKOddfRmeVO4KUsG9wM/branch/bQjlkhAqO2w3nk7NJF5EKS/Rainbow-BX?type=design&node-id=15871-3815&mode=design&t=9WMBduV6Zr4VpzTO-0

## What changed (plus any additional context for devs):
- Added the ability to unpair HW groups. Non-destructive, so a shorter, easier flow.

## What to test
- unpairing Ledger + Trezor hw groups and adding them back.

<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/05a59980-1266-47f8-824d-8b892b116013">


DanielSinclair (approved):
Tweaked the icon here. Functionality looks good!
Screenshot 2024-03-06 at 1 20 26 AM

src/entries/popup/pages/settings/walletsAndKeys/hardwareWalletWipePrompt.tsx
```diff
@@ -0,0 +1,174 @@
...
+  onClose: () => void;
+}) => {
+  const { currentAddress, setCurrentAddress } = useCurrentAddressStore();
+  const navigate = useRainbowNavigate();
+  const [wallet, setWallet] = useState<KeychainWallet | null>();
+
+  useEffect(() => {
+    const fetchWallets = async () => {
+      const walletFromKeychain = await getSettingWallets();
+      console.log(walletFromKeychain);
```
estebanmino, line 54: log

src/entries/popup/pages/settings/walletsAndKeys/hardwareWalletWipePrompt.tsx
```diff
@@ -0,0 +1,174 @@
...
+          await wipe();
+          navigate(ROUTES.WELCOME);
+          return;
+        }
+
+        const { hiddenWallets } = hiddenWalletsStore.getState();
+        const visibleWallets = allAccounts.filter((acc) => !hiddenWallets[acc]);
+        // if no more visible wallets, force unhide one and set as current
+        if (visibleWallets.length === 0) {
+          unhideWallet({ address: allAccounts[0] });
```
estebanmino, line 79: shouldn't this use hiddenWallets[0] instead?

BrodyHughes: I think in this case they would return the same thing. There may be an edge case where what you say needs to be done though. I think i would have to do something like:
```suggestion
          unhideWallet({
            address: Object.keys(hiddenWallets)[0] as `0x${string}`,
          });
```

BrodyHughes: but if visibleWallets.length === 0 then unhiding allAccounts[0] i think does the same thing. i may be confused tho.

estebanmino: that new change lgtm, is nice to maintain an order of names 👌  unhiding hidden wallets makes more sense than unhiding visible wallets

estebanmino (approved):

PR #1375, opened 2024-03-01
gridplus lattice 1 e2e test

BrodyHughes:
forked external PR to run CI/CD


e2e/helpers.ts
```diff
@@ -1069,7 +1116,7 @@ export async function delay(ms: number) {
 }
 
 export async function delayTime(
-  time: 'short' | 'medium' | 'long' | 'very-long',
+  time: 'short' | 'medium' | 'long' | 'very-long' | 'unbelievably-long',
```
BrodyHughes, line 1119:
```suggestion
  time: 'short' | 'medium' | 'long' | 'very-long',
```

BrodyHughes (commented):
did a quick lil onboarding test using mock data / IS_TESTING. Only issues I see:
- Wallets & Keys page uses the generic Ledger logo instead of the Grid Lattice + logo
- The first two pages of the flow don’t seem to match our other designs. mainly font size / spacing. this is just me nit picking tho. may need designer input.

Screenshot 2024-03-01 at 12 35 59 PM

BrodyHughes (commented):
**Finally got to test this with a physical GridPlus Lattice today:**
- Import flow worked well
- Eth mainnet swaps, sends + dapp interactions are the only thing that seemed to work for me. No L2s or sidechains. When attempting, the GridPlus screen would flash, then the browser extension would say ‘make sure your GridPlus is unlocked’
- Importing a duplicate wallet just caused the gridPlus import flow to get stuck on this screen:
Screenshot 2024-03-07 at 2 48 59 PM

- Also when you import a wallet with the GridPlus, then import another different wallet with it, they appear as separate wallet groups in our Wallets & Keys:
Screenshot 2024-03-07 at 2 49 28 PM

- Additionally, it would be nice if the flow was keyboard navigable. I have suggestions on how to make the first page of the flow work with our keyboard nav. You can add tabIndex={0} on most things to make it work, or in some cases, you can also use our Lens component.

PR #1379, opened 2024-03-05
add: blast

estebanmino:
Fixes BX-1337
Figma link (if any):

added missing params for migration and state migrations

you can check the ticket to repro and double check the fix works

image


src/core/types/chains.ts
```diff
@@ -20,6 +21,26 @@ export const chainHardhat: Chain = {
...
+  },
+  blockExplorers: {
+    default: { name: 'Blastscan', url: 'https://blastscan.io/' },
+  },
+  nativeCurrency: {
+    name: 'Blast',
+    symbol: 'BLAST',
+    decimals: 18,
+  },
+};
```
greg-schrammel, line 42: can't we update wagmi/viem? I think they got blast already

src/core/resources/assets/userAssets.ts
```diff
@@ -248,6 +248,7 @@ export async function parseUserAssets({
     (dict, currentChainId) => ({ ...dict, [currentChainId]: {} }),
     {},
   ) as ParsedAssetsDictByChain;
+  console.log('- parsedAssetsDict assets', assets);
```
greg-schrammel, line 251: logs?

src/core/types/chains.ts
dereknelson: i updated it & didn't see it

DanielSinclair: Should be in `2.7.18` and above

estebanmino: bumping

estebanmino: the new Chain object from viem is different than Chain object from wagmi 🙃

greg-schrammel (approved):

DanielSinclair:
Seeing an issue when building locally; my ENV does include the new Blast RPCs. Switch network RPC call seems to work fine.
Screenshot 2024-03-07 at 6 02 10 PM
Screenshot 2024-03-07 at 6 02 20 PM

src/core/types/chains.ts
jxom: Wat

estebanmino: https://github.com/rainbow-me/browser-extension/blob/master/package.json#L156

we're very late

estebanmino:
> Seeing an issue when building locally; my ENV does include the new Blast RPCs. Switch network RPC call seems to work fine. Screenshot 2024-03-07 at 6 02 10 PM Screenshot 2024-03-07 at 6 02 20 PM

@DanielSinclair  ok it should be ok now

DanielSinclair:
@estebanmino Tested again and seeing one last issue:
- Network appears as `Disabled` by default on a fresh install
Screenshot 2024-03-11 at 4 17 29 AM

BrodyHughes (approved):

PR #1380, opened 2024-03-05
polish swap review sheet

greg-schrammel:
Fixes BX-####
Figma link (if any):

- better format route part percentage
- fix fallback icon when protocol has no icon
- fixes a bug where if the route changed while you are in the review sheet sometimes the "routing via" row value vanished


derHowie (approved):
💅

src/entries/popup/components/ExternalImage/ExternalImage.tsx
```diff
@@ -137,9 +137,9 @@ const ExternalImage = (props: ExternalImageProps) => {
           ? { borderRadius: props.borderRadius }
           : {}),
         boxShadow: isLoading || error ? undefined : props.boxShadow,
-        height: props.height,
         overflow: error ? 'visible' : 'clip',
-        width: props.width,
+        height: props.height && (Number(props.height) ?? props.height),
+        width: props.width && (Number(props.width) ?? props.width),
```
estebanmino, line 142: i'm a bit confused, why is this?

estebanmino (approved):

src/entries/popup/components/ExternalImage/ExternalImage.tsx
greg-schrammel: when passing `height="20"` it didn't work but `height={20}` did, I think it expect to have a unit like `height="20px"` when it's a string (?)

DanielSinclair (approved):
Looks good to me!

PR #1383, opened 2024-03-06
fix: blast custom rpc

estebanmino:
Fixes BX-1337
Figma link (if any):

fixed chain ids and unique ids of native asset that was causing this issue when getting assets from backend originating the issue

image

### DIFF
--- a/src/core/resources/assets/customNetworkAssets.ts
+++ b/src/core/resources/assets/customNetworkAssets.ts
@@ -220,7 +220,7 @@ async function customNetworkAssetsFunction({
                   isNativeAsset: true,
                   name: chain.nativeCurrency.symbol,
                   symbol: chain.nativeCurrency.symbol,
-                  uniqueId: `${nativeAssetAddress}_${chain.id}`,
+                  uniqueId: `${'eth'}_${chain.id}`,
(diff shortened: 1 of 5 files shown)


src/core/resources/assets/customNetworkAssets.ts
```diff
@@ -220,7 +220,7 @@ async function customNetworkAssetsFunction({
                   isNativeAsset: true,
                   name: chain.nativeCurrency.symbol,
                   symbol: chain.nativeCurrency.symbol,
-                  uniqueId: `${nativeAssetAddress}_${chain.id}`,
+                  uniqueId: `${'eth'}_${chain.id}`,
```
greg-schrammel, line 223: is this 'eth' here right?

greg-schrammel (approved):

derHowie (approved):

src/core/resources/assets/customNetworkAssets.ts
estebanmino: yeah, uniqueid for native assets is eth otherwise i can't filter these out of the token list and you'll see duplicated native assets for this

PR #1384, opened 2024-03-06
Activity details fixes

greg-schrammel:
Fixes BX-1338
Figma link (if any):

### DIFF
--- a/src/core/resources/transactions/transaction.ts
+++ b/src/core/resources/transactions/transaction.ts
@@ -34,12 +34,11 @@ export const fetchTransaction = async ({
   currency,
   chainId,
 }: {
-  hash?: TxHash;
+  hash: TxHash;
   address: Address;
   currency: SupportedCurrencyKey;
-  chainId?: ChainId;
+  chainId: ChainId;
 }) => {
-  if (!chainId || !hash) return undefined;
   try {
     const response = await addysHttp.get<{
       payload: { transaction: TransactionApiResponse };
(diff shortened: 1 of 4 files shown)


src/core/resources/transactions/transaction.ts
```diff
@@ -122,13 +121,13 @@ const getCustomChainTransaction = async ({
...
-  hash?: Hash;
+  chainId: number;
+  hash: Hash;
 }) => {
-  if (!chainId || !hash) return undefined;
   const provider = getProvider({ chainId });
   const transaction = await provider.getTransaction(hash);
-  if (!transaction) return undefined;
+  if (!transaction)
+    throw `getCustomChainTransaction: couldn't find transaction`;
```
greg-schrammel, line 130: react-query don't it like when a `queryFn` returns `undefined`

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -248,7 +248,7 @@ function NetworkData({ transaction: tx }: { transaction: RainbowTransaction }) {
           </Inline>
         }
       />
-      <FeeData transaction={tx} />
+      {tx.status != 'pending' && <FeeData transaction={tx} />}
```
greg-schrammel, line 251: I'm working with "stuck" transactions and it's weird to see the fee data rows being in a perpetual skeleton state when it's not even fetching it, but happy to revert if intended

Issue #1382, opened 2024-03-06
wallet_watchAsset throws an error


navFooh:
Using wagmi `2.6.x`

```typescript
import { getConnectorClient } from '@wagmi/core'

const client = await getConnectorClient({ ... }, {
  account: '0x...',
  chainId: 314159,
  connector: ...,
})

await client.request({
  method: 'wallet_watchAsset',
  params: {
    type: 'ERC20',
    options: {
      address: '0xaC26a4Ab9cF2A8c5DBaB6fb4351ec0F4b07356c4',
      symbol: 'WFIL',
      decimals: 18,
    },
  },
})
```

When sending the `wallet_watchAsset` request, a window pops up to approve adding the token:

<img width="261" alt="image" src="https://github.com/rainbow-me/rainbow/assets/394452/4997826a-43f5-46dd-ab18-f1744d164a43">

However, when clicking Approve, an error is thrown: `User rejected the request.`

PR #1387, opened 2024-03-08
fix: BX-1341

estebanmino:
Fixes BX-####
Figma link (if any):

fixing ticket and returning true when request is accepted

https://github.com/rainbow-me/browser-extension/issues/1382

Fixes https://github.com/rainbow-me/browser-extension/issues/1382


src/entries/popup/pages/messages/WatchAsset/index.tsx
```diff
@@ -165,7 +164,7 @@ export const WatchAsset = ({
         });
       }
 
-      approveRequest(null);
+      approveRequest(true);
```
greg-schrammel, line 167: looking at `ApproveAppRequest.tsx` seems like `approveRequest` and `rejectRequest` send the same msg to background but if the payload is `null` it considers a reject, shouldn't we solve it in that component? 
I think `AddEthereumChain.tsx` also calls `approveRequest(null)`

estebanmino: yeah that's exactly the case, so we're sending true instead

maybe your AddEthereumChain is outdated https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/pages/messages/AddEthereumChain/index.tsx#L79

greg-schrammel (approved):

derHowie (approved):

PR #1398, opened 2024-03-14
[BX-1139] support enter/return keypress to submit "save contact" form (Round 2)

derHowie:
Supporting enter to submit on save contact


src/core/providers/proxy.ts
```diff
@@ -4,9 +4,9 @@ export const proxyRpcEndpoint = (endpoint: string, chainId: ChainId) => {
   if (
     endpoint !== 'http://127.0.0.1:8545' &&
     endpoint !== 'http://localhost:8545' &&
-    !endpoint.includes('http://10.') &&
-    !endpoint.includes('http://192.168') &&
-    !endpoint.match(/http:\/\/172.(1[6-9]|2[0-9]|3[0-1])./)
+    !endpoint?.includes('http://10.') &&
+    !endpoint?.includes('http://192.168') &&
+    !endpoint?.match(/http:\/\/172.(1[6-9]|2[0-9]|3[0-1])./)
```
estebanmino, line 9: i'm wondering if there's a bug happening because of this and typecheck would just fix the crash

https://github.com/rainbow-me/browser-extension/blob/master/src/core/wagmi/createWagmiClient.ts#L51

we're checking whether the original rpc is present or not, maybe we have some bad params for a chain?

derHowie: In this case `getOriginalRpcEndpoint` is returning `{ http: undefined }` for blast.

BrodyHughes (approved):

estebanmino (approved):

PR #1402, opened 2024-03-14
feat: hide / unhide tokens

magiziz:
Fixes BX-1356, BX-1365

1. Context menu UI in the token list is now sorted by the right order. Also added keyboard shortcut for pin / unpin and hide / unhide.
2. Token details more menu is also sorted by the right order, but without shortcuts.
3. Added translations for hide, unhide and view on coingecko.
4. Added shorcuts to pin / unpin, hide /unhide and copy address + tracking (analytics)

https://github.com/rainbow-me/browser-extension/assets/53529533/5b2ced74-cb8e-4931-a8f0-b93e5eca62ee

## What to test

1. In the token list right click on any token and you should see a overflow menu with option "Hide x token".
5. Hidden tokens should be hidden from token list, swap, send and bridge, but you should be able to unhide the token in token details page and token shouldn't be hidden when you do `cmd + K`.
6. For some tokens like testnet or custom network ones, the dropdown is usually hidden in the details page, but if you hide  testnet or custom network tokens the dropdown should appear to unhide the token.
7. Try to also pin a token, then hide it. It should also trigger unpin behind the scenes. You'll see it ones you unhide the token in the token details page, then go back to token list section in the homepage.


src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
```diff
@@ -162,6 +164,19 @@ export function TokenContextMenu({ children, token }: TokenContextMenuProps) {
                 })}
           </TextOverflow>
         </ContextMenuItem>
+        <ContextMenuItem
+          symbolLeft="eye.slash.fill"
+          onSelect={() => {
+            addHiddenAsset(token.address, token.chainId);
```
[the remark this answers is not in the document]
magiziz: Great idea 💡, let me do that 🙏

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -404,6 +434,8 @@ export function TokenDetails() {
     isTestnetChainId({ chainId: token?.chainId }) || !!customAsset
   );
 
+  const hidden = isHidden(token.address, token.chainId);
```
estebanmino, line 437: same here, isHidden could receive token as param and then inside of it you can use token.uniqueId

estebanmino (approved):
🪨

magiziz:
For anyone who is reviewing this you might see that i don't use `uniqueId` to store hidden assets, but instead we store token address with chain id (`${asset.address}-${asset.chainId}`). This is because swap page has different `uniqueId`'s than token list page. Works for now, but might consider a follow up refactor PR once this is fixed in the backend.

src/entries/popup/hooks/useHiddenAssets.ts
```diff
@@ -0,0 +1,41 @@
+import { useCallback } from 'react';
+
+import { useHiddenAssetStore } from '~/core/state/hiddenAssets/hiddenAssets';
+import { ParsedUserAsset } from '~/core/types/assets';
+import { SearchAsset } from '~/core/types/search';
+
+export const useHiddenAssets = () => {
```
greg-schrammel, line 7: why are these methods not in the store itself?

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -284,39 +294,58 @@ function MoreOptions({ token }: { token: ParsedUserAsset }) {
...
       <DropdownMenuContent align="end">
-        <AccentColorProvider
-          color={token.colors?.primary || token.colors?.fallback}
-        >
-          {!isNative && (
-            <DropdownMenuItem
-              symbolLeft="doc.on.doc.fill"
-              onSelect={() => copyAddress(token.address)}
+        {swappable && (
+          <Fragment>
```
greg-schrammel, line 298: if AccentColorProvider wraps it all do we need the fragment?

src/entries/popup/hooks/useHiddenAssets.ts
magiziz: Yeah. We could do it from store itself, but there would be a lot of duplicate functions since we use the same `${asset.address}-${asset.chainId}` string to add and remove hidden assets + `isHidden` is used a lot so didn't want to duplicate that as well. 

Keeping this in a hook would just make things simpler. Lmk what you think 🙏

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
magiziz: Oohh great catch 👀. Fixed it in this [commit](https://github.com/rainbow-me/browser-extension/pull/1402/commits/85a314b3b5f1cb1c8a7c1be6cabe738f41ac7909).

src/entries/popup/hooks/useHiddenAssets.ts
greg-schrammel: I mean something like 
```ts
export const hiddenAssets = createStore<HiddenAssetState>(
  (set, get) => ({
    hiddenAssets: [],
    addHiddenAsset: (asset: { chainId: ChainId, address: Address }) => {
      const { hiddenAssets } = get();
      const uniqueId = `${asset.address}-${asset.chainId}`
      set({ hiddenAssets: [...hiddenAssets, uniqueId] });
    },
    isHidden: (asset) => {
      const { hiddenAssets } = get();
      const uniqueId = `${asset.address}-${asset.chainId}`
      return hiddenAssets.includes(uniqueId)
    }
```
so whats outside the store don't need to know about how it stores its data, just pass an asset and it handles the rest

greg-schrammel: and basically rename `useHiddenAssetStore` -> `useHiddenAssets` I think would work the same

greg-schrammel (approved):

src/entries/popup/hooks/useHiddenAssets.ts
magiziz: Ah ok. Yeah makes sense now, let me do that. Thanks 🙏

DanielSinclair:
@magiziz Before you merge, going to send some tweaks to the context menu orders. Looks like a few regressions over time there.

static/json/languages/en_US.json
```diff
@@ -1458,14 +1458,17 @@
     "approvals": "Token Approvals",
     "this_token_is_on_network": "This %{symbol} is on the %{chainName} network",
     "view_on": "View on %{explorer}",
+    "view_on_coingecko": "View on CoinGecko",
```
magiziz, line 1461: Didnt want to use `view_on` since that's related to explorer's only

DanielSinclair (changes requested):
cc @greg-schrammel Will need another review early next week after some bug fixes

lavamoat/build-webpack/policy.json
DanielSinclair, line 1: Some strange policy changes here; might want to drop. Maybe an issue with your styling plugin?

src/entries/popup/pages/home/TokenDetails/KeyboardShortcutListener.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const condition = useCallback(() => true, []);
+
+  const handleContextMenuShortcuts = useCallback(
+    (e: KeyboardEvent) => {
+      switch (e.key) {
+        case shortcuts.tokens.PIN_ASSET.key:
+          trackShortcut({
+            key: shortcuts.tokens.PIN_ASSET.display,
+            type: 'tokens.pin',
```
DanielSinclair, line 28: These types would be different between Tokens (Home) and Token Details. Should create a subset type for token details shortcuts

src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
```diff
@@ -155,13 +181,52 @@ export function TokenContextMenu({ children, token }: TokenContextMenuProps) {
...
+              {i18n.t('token_details.more_options.copy_address')}
+            </Text>
+            <Text size="11pt" color="labelTertiary" weight="medium">
+              {truncateAddress(token.address)}
+            </Text>
+          </ContextMenuItem>
+        )}
+        {explorer && (
+          <Box style={{ margin: '4px 0' }}>
+            <Separator />
```
DanielSinclair, line 218: We have a `ContextMenuSeparator` type for this behavior; should use that

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -287,34 +304,86 @@ function MoreOptions({ token }: { token: ParsedUserAsset }) {
...
+                  <Text size="14pt" weight="semibold">
+                    {i18n.t('token_details.more_options.copy_address')}
+                  </Text>
+                  <Text size="11pt" color="labelTertiary" weight="medium">
+                    {truncateAddress(token.address)}
+                  </Text>
+                </DropdownMenuItem>
+              )}
+              <Box style={{ margin: '4px 0' }}>
+                <Separator />
```
DanielSinclair, line 368: Also have a `<DropdownMenuSeperator>` component for this

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -287,34 +304,86 @@ function MoreOptions({ token }: { token: ParsedUserAsset }) {
...
+              symbolLeft="pin.fill"
+              onSelect={() => {
+                if (pinned) {
+                  removedPinnedAsset({ uniqueId: token.uniqueId });
+                  return;
+                }
+
+                addPinnedAsset({ uniqueId: token.uniqueId });
+              }}
             >
```
DanielSinclair, line 318: Looks like the shortcut hint UI is missing here 
Screenshot 2024-03-23 at 3 29 49 AM

src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
```diff
@@ -76,17 +92,55 @@ export function TokenContextMenu({ children, token }: TokenContextMenuProps) {
...
 
   return (
     <DetailsMenuWrapper closed={true} onOpenChange={onOpenChange}>
       <ContextMenuTrigger asChild>{children}</ContextMenuTrigger>
       <ContextMenuContent>
+        <KeyboardShortcutListener
+          hideToken={hideToken}
+          togglePin={togglePin}
+          copyTokenAddress={copyTokenAddress}
+        />
```
DanielSinclair, line 143: Have a look at how we do this for the tokens screen (home) with the `useTokensShortcuts` hook. I think we want to continue with consistency of that approach rather than a wrapper
https://github.com/rainbow-me/browser-extension/blob/764e5153b9c746cb7cf20f79ee46d6e103df3ec8/src/entries/popup/hooks/useTokensShortcuts.ts#L23

static/json/languages/en_US.json
```diff
@@ -1458,14 +1458,18 @@
...
       "copy_address": "Copy Address",
       "view_on_explorer": "View on %{explorer}",
       "hide": "Hide Token",
       "report": "Report Token",
       "unpin_token": "Unpin %{name}",
-      "pin_token": "Pin %{name}"
+      "pin_token": "Pin %{name}",
+      "hide_token": "Hide %{name}",
+      "unhide_token": "Unhide %{name}",
+      "hide_token_toast": "%{name} hidden"
```
DanielSinclair, line 1472: Check out the `hide_balances_toast` example; I think we want to bump this out of this section and have a separate toast key for these actions. Try this for a string: `Your %{name} is hidden`. Would also need a toast for unhide and pin/unpin I think. I realized after my prior comment that we do want a toast on the Token Details page for those, but don't need one on Tokens (because of the visual list shift)

src/entries/popup/hooks/useHiddenAssets.ts
DanielSinclair, line 1: Don't think we need this file. Generally we use hooks for read-only state (i.e. the full list of wallets, or memoized versions of that list), whereas these are duplicate mutations already represented in `useHiddenAssetStore`. Should probably use that directly, and then perform the `isHidden` check inline in the components based on the results of `hiddenAssets`

src/core/state/hiddenAssets/hiddenAssets.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type UpdateHiddenAssetFn = ({ uniqueId }: UpdateHiddenAssetArgs) => void;
+
+export interface HiddenAssetState {
+  hiddenAssets: string[];
+  addHiddenAsset: UpdateHiddenAssetFn;
+  removeHiddenAsset: UpdateHiddenAssetFn;
+}
+
+export const hiddenAssets = createStore<HiddenAssetState>(
```
DanielSinclair, line 17: `hiddenAssets` should be `hiddenAssetsStore` here

DanielSinclair: If we needed a hidden assets hook at some point (like to display a list of just the hidden assets) we wouldn't want the name collision

src/core/state/hiddenAssets/hiddenAssets.ts
```diff
@@ -0,0 +1,52 @@
...
+  hiddenAssets: string[];
+  addHiddenAsset: UpdateHiddenAssetFn;
+  removeHiddenAsset: UpdateHiddenAssetFn;
+}
+
+export const computeUniqueIdForHiddenAsset = (
+  asset: ParsedUserAsset | SearchAsset,
+) => {
+  return `${asset.address}-${asset.chainId}`;
+};
```
derHowie, line 24: Is there a particular reason you're building this yourself? Now that https://github.com/rainbow-me/browser-extension/commit/b1e3f5edc7d328978f07338eab49025b372c5f28 is merged, would it make more sense to grab this from the asset itself? Haven't fully thought this through, but would be curious to hear your reasoning.

src/entries/popup/pages/home/Tokens.tsx
```diff
@@ -144,17 +159,18 @@ export function Tokens() {
...
       pinnedAssets.some(({ uniqueId }) => uniqueId === assetUniqueId),
     [pinnedAssets],
   );
 
+  const combinedFilteredAssets = useMemo(
+    () =>
+      [...assets, ...customNetworkAssets].filter((asset) => !isHidden(asset)),
+    [assets, customNetworkAssets, isHidden],
+  );
+
```
derHowie, line 173: The naming in this file is getting kind of wonky. We are creating `combinedFilteredAssets` then using that value to create `filteredAssets`.

Maybe we could describe the filter happening in the first value so it's easier to think about later. Something like `unhiddenAssets` or `assetsToDisplay` or `assetsToShow`?

src/entries/popup/pages/send/index.tsx
```diff
@@ -116,6 +134,11 @@ export function Send() {
     sortMethod,
   } = useSendAsset();
 
+  const filteredNonHiddenAssets = useMemo(
```
derHowie, line 137: Whatever name you choose for the above comment, i'd mirror that here since we're doing the same operation.

src/entries/popup/pages/home/Tokens.tsx
```diff
@@ -285,7 +301,11 @@ export function Tokens() {
                 style={{ height: size, y: start }}
               >
                 {pinned && <TokenMarkedHighlighter />}
-                <TokenRow token={token} testId={`coin-row-item-${index}`} />
+                <TokenRow
+                  containerRef={containerRef}
```
derHowie, line 305: My guess is that you're passing this all the way down to the context menus so that if we use a shortcut, we can simulate a click on the main container to close the context menu?

Do you think there'd be negative performance implications in just using this hook in the context menu itself rather than prop drilling it down two levels?

derHowie (commented):
Looks good, just a couple questions.

src/entries/popup/pages/home/TokenDetails/TokenContextMenu.tsx
```diff
@@ -76,17 +96,54 @@ export function TokenContextMenu({ children, token }: TokenContextMenuProps) {
...
 
   return (
     <DetailsMenuWrapper closed={true} onOpenChange={onOpenChange}>
       <ContextMenuTrigger asChild>{children}</ContextMenuTrigger>
       <ContextMenuContent>
+        <TokenMenuShortcutListener
+          hideToken={hideToken}
+          pinToken={togglePinToken}
+          copyTokenAddress={copyTokenAddress}
+        />
```
derHowie, line 146: Do you think this functionality could live within `useTokenShortcuts` with the other token shortcut handlers?

magiziz: Yeah was thinking of doing that too, but realised we already have it as a global scope [here](https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/pages/home/Tokens.tsx#L229) in the homepage. I could take the hook and listen to it again, but we would face duplications which would cause double tracking.

src/entries/popup/pages/home/Tokens.tsx
magiziz: Good point. Fixing rn

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
magiziz: We're going to fix this in another ticket. Seems like there is a bug with `<DropdownItem />`

derHowie (approved):
LGTM, thanks for the changes

src/entries/popup/hooks/useTokensShortcuts.ts
```diff
@@ -28,17 +38,80 @@ export function useTokensShortcuts() {
...
+        simulateClick(containerRef.current);
+        return;
+      }
+      addPinnedAsset({ uniqueId: _selectedToken.uniqueId });
+      triggerToast({
+        title: i18n.t('token_details.toast.pin_token', {
+          name: _selectedToken.symbol,
+        }),
+      });
+      simulateClick(containerRef.current);
```
greg-schrammel, line 97: are these simulateClick on the container to close the context menu?

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -287,34 +342,66 @@ function MoreOptions({ token }: { token: ParsedUserAsset }) {
...
+                  <Text size="11pt" color="labelTertiary" weight="medium">
+                    {truncateAddress(token.address)}
+                  </Text>
+                </DropdownMenuItem>
+              )}
+              <DropdownMenuSeparator />
+              <DropdownMenuItem
+                symbolLeft="safari"
+                external
+                onSelect={() => window.open(getCoingeckoUrl(token), '_blank')}
```
greg-schrammel, line 391: we have `goToNewTab` function somewhere, not sure why 
@estebanmino is there a reason to use `chrome.tabs.create` instead of `window.open`?

src/entries/popup/pages/home/TokenDetails/TokenDetails.tsx
```diff
@@ -267,9 +273,58 @@ export const getCoingeckoUrl = ({
...
+  }, [
+    token,
+    hidden,
+    pinned,
+    removedPinnedAsset,
+    addHiddenAsset,
+    removeHiddenAsset,
+  ]);
+
+  const togglePinToken = useCallback(() => {
```
greg-schrammel, line 320: if your not passing this function to a memoized component it don't need keep it's ref stable (`useCallback`)

greg-schrammel: because like it is recreating the function anyway, just going thru more effort to check the cache and return the previous if deps didn't change
I mean `useCallback` is super cheap but not free

not a change request, just a note

src/entries/popup/hooks/useTokensShortcuts.ts
magiziz: Yep

greg-schrammel (approved):

DanielSinclair (approved):

PR #1405, opened 2024-03-14
fix: BX-1350

estebanmino:
Fixes BX-1350
Figma link (if any):

last week fixing blast i made this change https://github.com/rainbow-me/browser-extension/pull/1383/files#diff-3b6eb6e4c37dcc8c725f157d7953e3de4245db86d2724ccd45939fba1793bd62R223 to emulate what backend returns as uniqueId for native asset

that was correct but i didn't change it in the method that gives us the native asset while estimating gas https://github.com/rainbow-me/browser-extension/pull/1405/files#diff-4fd4cdb2833891791c48304800a1085b7c1c170b44e79c2121a5e3d701f7ce65R65 so native asset was undefined for custom networks, causing the issue

I updated all instances of this to have custom native tokens in the same format as backend

### DIFF
--- a/src/entries/popup/hooks/useVisibleTokenCount.ts
+++ b/src/entries/popup/hooks/useVisibleTokenCount.ts
@@ -49,6 +49,9 @@ export const useVisibleTokenCount = () => {
     },
   );
 
+  console.log('assets', assets);
+  console.log('customNetworkAssets', customNetworkAssets);
+
   const allAssets = useMemo(
     () =>
       uniqBy(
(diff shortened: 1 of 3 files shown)


src/entries/popup/hooks/useVisibleTokenCount.ts
```diff
@@ -49,6 +49,9 @@ export const useVisibleTokenCount = () => {
     },
   );
 
+  console.log('assets', assets);
+  console.log('customNetworkAssets', customNetworkAssets);
```
BrodyHughes, line 53: logs

derHowie (approved):
🚀

DanielSinclair (approved):

PR #1413, opened 2024-03-18
fix speed up/cancel transaction

greg-schrammel:
Fixes BX-####
Figma link (if any):

- fixes the activity tab while speeding up/cancelling a transaction
- fixes the crash when trying to speed up

## What to test

try to speed up and cancel a transaction


src/entries/popup/handlers/wallet.ts
```diff
@@ -157,13 +158,33 @@ export const sendTransaction = async (
...
       'send_transaction',
       params,
-    ) as unknown as TransactionResponse;
+    );
+
+    return deserializeBigNumbers(transactionResponse);
   }
 };
 
+function deserializeBigNumbers<T>(obj: T) {
```
greg-schrammel, line 170: not 100% sure about this solution, open to suggestions
but `walletAction` bridges to the background worker and back, in this back and forth it serializes BigNumbers as `{ type: 'BigNumber', _hex: "0x..." }`
so sometimes we got an `[object Object]` instead of the bignumber in some places

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -288,13 +293,40 @@ function NetworkData({ transaction: tx }: { transaction: RainbowTransaction }) {
...
+  return (
+    <ExplainerSheet
+      show={explainer === 'speed_up_error'}
+      onClickOutside={() => setSearchParams({})}
+      header={{
+        icon: <Symbol symbol="xmark.circle.fill" color="red" size={32} />,
+      }}
+      title="Failed to speed up transaction"
+      description={[
+        'A speed up can fail when the state of the network changed since you first submitted the transaction',
```
greg-schrammel, line 309: @DanielSinclair let's define a copy for this and I add to the translations

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -545,6 +577,8 @@ export function ActivityDetails() {
   });
   const navigate = useRainbowNavigate();
 
+  console.log(transaction);
```
DanielSinclair: log

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
DanielSinclair: @greg-schrammel Some copy:

"Speed up failed"

"Attempts to speed up a transaction can fail during network congestion, gas limit increases, or when the original transaction is confirmed first"

src/entries/popup/handlers/wallet.ts
greg-schrammel: I started to do it and got afraid of changing the type and breaking in unexpected places, because it's being casted as `TransactionResponse` and that is why we didn't caught it for so long to begin with
to do it without changing the type, I'd have to then deserialize by hand like 
```ts
return {
 ...response,
 maxBaseFee: BigNumber.from(response.maxBaseFee)
 ...
}
```
which is more error prone and effort

another idea is to do this generic BigNumber deserialization inside `walletAction` so we don't have to ever think about it again in other places

estebanmino: okok sounds good, can we move this to a util file somewhere else?

estebanmino (approved):

estebanmino:
bumping this small comment https://github.com/rainbow-me/browser-extension/pull/1413#discussion_r1581412035

src/entries/popup/handlers/wallet.ts
derHowie: +1

derHowie (approved):
🚢

PR #1419, opened 2024-03-21
Activities fixes

greg-schrammel:
Fixes BX-####
Figma link (if any):

- show right explorer (like socketscan for bridges) on context click at a tx in the activities list
- fix a bug that was hiding failed txs from the list
- it was trying to fetch pending txs as if they were from custom networks and overriding its data, so like a pending swap was gonna display as a "send" in the next refetch
- polishes the activity details sheet values

### DIFF
--- a/src/core/resources/transactions/consolidatedTransactions.ts
+++ b/src/core/resources/transactions/consolidatedTransactions.ts
@@ -136,7 +136,7 @@ async function parseConsolidatedTransactions(
   currency: SupportedCurrencyKey,
 ) {
   const data = message?.payload?.transactions || [];
-  const parsedTransactionPromises = data
+  return data
     .map((tx) =>
       parseTransaction({
         tx,
@@ -145,11 +145,6 @@ async function parseConsolidatedTransactions(
       }),
     )
     .filter(Boolean);
-
-  const parsedConsolidatedTransactions = (
-    await Promise.all(parsedTransactionPromises)
-  ).flat();
-  return parsedConsolidatedTransactions;
 }
 
 // ///////////////////////////////////////////////
--- a/src/core/resources/transactions/transaction.ts
+++ b/src/core/resources/transactions/transaction.ts
@@ -1,3 +1,4 @@
+import { formatUnits } from '@ethersproject/units';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { Hash, getProvider } from '@wagmi/core';
 import { Address } from 'wagmi';
@@ -10,7 +11,11 @@ import {
   consolidatedTransactionsQueryFunction,
   consolidatedTransactionsQueryKey,
 } from '~/core/resources/transactions/consolidatedTransactions';
-import { useCurrentAddressStore, useCurrentCurrencyStore } from '~/core/state';
+import {
+  pendingTransactionsStore,
+  useCurrentAddressStore,
+  useCurrentCurrencyStore,
+} from '~/core/state';
 import { useTestnetModeStore } from '~/core/state/currentSettings/testnetMode';
 import { ChainId } from '~/core/types/chains';
 import {
@@ -27,6 +32,14 @@ import { RainbowError, logger } from '~/logger';
 type ConsolidatedTransactionsResult = QueryFunctionResult<
   typeof consolidatedTransactionsQueryFunction
 >;
+
+const searchInLocalPendingTransactions = (userAddress: Address, hash: Hash) => {
+  const { pendingTransactions } = pendingTransactionsStore.getState();
+  const localPendingTx = pendingTransactions[userAddress]?.find(
+    (tx) => tx.hash === hash,
+  );
+  return localPendingTx;
+};
 
 export const fetchTransaction = async ({
   hash,
@@ -48,6 +61,9 @@ export const fetchTransaction = async ({
     });
     const tx = response.data.payload.transaction;
     if (response.data.meta.status === 'pending') {
+      const localPendingTx = searchInLocalPendingTransactions(address, hash);
+      if (localPendingTx) return localPendingTx;
+
       const providerTx = await getCustomChainTransaction({ chainId, hash });
       return providerTx;
     }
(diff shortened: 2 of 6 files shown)


src/core/resources/transactions/consolidatedTransactions.ts
```diff
@@ -145,11 +145,6 @@ async function parseConsolidatedTransactions(
       }),
     )
     .filter(Boolean);
-
-  const parsedConsolidatedTransactions = (
-    await Promise.all(parsedTransactionPromises)
```
greg-schrammel, line 150: parseTransaction is not a promise

estebanmino: we have the same here too https://github.com/rainbow-me/browser-extension/blob/master/src/core/resources/transactions/transactions.ts#L119

estebanmino: would be nice to remove as well

src/core/resources/transactions/transaction.ts
```diff
@@ -48,16 +61,25 @@ export const fetchTransaction = async ({
     });
     const tx = response.data.payload.transaction;
     if (response.data.meta.status === 'pending') {
+      const localPendingTx = searchInLocalPendingTransactions(address, hash);
```
estebanmino, line 64: why do we want to do this?

i'm thinking that we need an updated state of the tx if BE has it as pending

src/core/resources/transactions/transaction.ts
```diff
@@ -146,7 +172,7 @@ const getCustomChainTransaction = async ({
         from: transaction.from as Address,
         to: transaction.to as Address,
         data: transaction.data,
-        value: transaction.value.toString(),
+        value,
```
estebanmino, line 175: i don't think we want to do this, i assume that value is in wei

is not breaking other places of the extension?

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -153,29 +154,41 @@ const InfoValueSkeleton = () => <Skeleton width="50px" height="12px" />;
...
+      BigNumber.from(tx.maxFeePerGas).sub(tx.maxPriorityFeePerGas).toString());
 
-  const gasPrice = tx.gasPrice && formatUnits(tx.gasPrice, 'gwei');
+  let fee;
+  if (native !== undefined && native.fee !== undefined) {
+    // if the fee is less than $0.01, the provider returns 0 so we display it as <$0.01
+    const feeInNative = +native.fee <= 0.01 ? 0.01 : native.fee;
+    fee = `${+feeInNative <= 0.01 ? '<' : ''}${formatCurrency(feeInNative)}`;
+  } else {
+    // handle custom networks fee
```
estebanmino, line 171: is this finished?

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -217,24 +225,25 @@ function FeeData({ transaction: tx }: { transaction: RainbowTransaction }) {
...
   const { rainbowChains } = useRainbowChains();
   const chain = getChain({ chainId: tx.chainId });
 
+  const formattedValueInNative =
+    Number(native.value) > 0 && formatCurrency(native.value);
+  const formattedValue =
+    Number(tx.value) > 0 &&
+    `${formatNumber(tx.value)} ${chain.nativeCurrency.symbol}`;
+
+  const value = formattedValueInNative || formattedValue;
```
estebanmino, line 238: would be nice to have this in a method

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
greg-schrammel: yeah not sure how to do that yet, left a comment for later

src/core/resources/transactions/transaction.ts
greg-schrammel: backend doesn't return all the tx data when it's pending, and the `getCustomChainTransaction` overrides the transaction type to `send` as it can't know that metadata
I don't see what new data we could grab from a pending tx like when it completes/reverts we update

src/core/resources/transactions/transaction.ts
greg-schrammel: `parseTransaction` formats the value like that, is the customChainTransaction used differently then a normal supported transaction?
didn't notice it breaking anything but possible

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
```diff
@@ -153,29 +154,41 @@ const InfoValueSkeleton = () => <Skeleton width="50px" height="12px" />;
...
+  let fee;
+  if (native !== undefined && native.fee !== undefined) {
+    // if the fee is less than $0.01, the provider returns 0 so we display it as <$0.01
+    const feeInNative = +native.fee <= 0.01 ? 0.01 : native.fee;
+    fee = `${+feeInNative <= 0.01 ? '<' : ''}${formatCurrency(feeInNative)}`;
+  } else {
+    // handle custom networks fee
+  }
+
+  if ((!baseFee || !tx.maxPriorityFeePerGas) && !tx.gasPrice) return null;
```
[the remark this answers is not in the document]
greg-schrammel: what should we call the method? just tried but it seemed more confusing to me, not sure how to abstract and not make it harder to understand whats goin on

src/core/resources/transactions/transaction.ts
estebanmino: ok went through the code and this is only fetching tx from be i was thinking this could be used for custom networks txs

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
estebanmino: so we won't merge this yet?

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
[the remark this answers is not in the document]
estebanmino: not sure how to call it lol but is odd to have number calculations there

src/core/resources/transactions/transaction.ts
estebanmino: right

image

i don't get why is working then 🤔

greg-schrammel: `providerTx` is early returning before getting to `parseTransaction`

src/entries/popup/pages/home/Activity/ActivityDetails.tsx
greg-schrammel: figured it out 🎉

src/core/providers/proxy.ts
```diff
@@ -2,6 +2,7 @@ import { ChainId } from '../types/chains';
 
 export const proxyRpcEndpoint = (endpoint: string, chainId: ChainId) => {
   if (
+    endpoint &&
```
estebanmino, line 5: i think i saw other pr trying to fix this before? wasn't fixed?

PR #1425, opened 2024-03-22
chore: report nft shortcuts, nft context menu fixes and Cmd+K bug fixes

DanielSinclair:
Fixes BX-1441, BX-1440, BX-1444
Figma link (if any):

- NFT shortcut menu doesn't close when interacting with a shortcut (NFT tab in homepage)
- If a context menu is closed after interaction and you try to click the "C" shortcut, it won't copy user's address, but instead copies token id address still.
- When displaying NFT's in "grid" mode the `Cmd+K` doesn't work if you hide the last NFT for the section.
- Added report NFT feature with hide functionality

Check bug videos below 👇

https://github.com/rainbow-me/browser-extension/assets/53529533/3a4244e6-d63c-40c6-9a80-3430bcbd1ee8

https://github.com/rainbow-me/browser-extension/assets/53529533/41383c52-d153-472a-9dee-2d9d3a354ba8

## What to test

- Report NFT feature in NFT tab (homepage and NFT detail page)
- Make sure shortcuts closes context menu if you start interacting with shortcuts (homepage and NFT detail page)
- Make sure `Cmd+K` works after your hide the last NFT when using "grid" section mode
- Make sure when context menu closes, after you click "C" shortcut it should copy user's address instead of NFT address
- Test the "R" shortcut for refreshing NFTs and when reporting an NFT through context menu. Both should work


DanielSinclair:
cc @magiziz Looks like we have a merge conflict here

DanielSinclair:
@magiziz Fixes are looking good, but spotted a few other things:
- Noticed that the `cmd` key seems to dismiss the NFT Details context menu (noticed while taking a screenshot)
- The cursor for the Copy item in the NFT Details was originally associated with the Download button, but I think we can remove it
- There is also a different cursor for the Download button in both NFT context menus (indicating download link); can we default that to the normal pointer cursor?
- I noticed the Download button in NFT Details sometimes/always triggers 2 downloads when clicking
- Looks like we're missing a Toast when using the Hide shortcut on the NFT Gallery

static/json/languages/en_US.json
```diff
@@ -1351,7 +1351,9 @@
       "view_gallery": "View Gallery"
     },
     "toast": {
-      "spam_reported": "Your report has been received!"
+      "spam_reported": "Your report has been received!",
+      "hidden": "NFT hidden",
+      "unhidden": "NFT unhidden"
```
magiziz, line 1356: Lmk if there are any better names we can go for here. Currently some NFT names are very long so wanted to simplify it.

BrodyHughes (commented):
Core functionality all works for me.

A few thoughts:
- hiding all NFTs leaves you with a blank page instead of the placeholder page that says "your nfts will show up here"
- using 'r' for refresh and 'r' for report seems a little odd. maybe we need a 'are you sure you want to report this?' the action seems somewhat destructive bc i couldn't find a way to 'un-report' or 'un-hide' an NFT

static/json/languages/en_US.json
```diff
@@ -1306,6 +1306,8 @@
   "nfts": {
     "empty_state_header": "NFTs",
     "empty_state_description": "Your NFTs will appear here!",
+    "report_nft_action_text": "Report",
+    "report_nft_confirm_description": "Confirm report of NFT?",
```
BrodyHughes, line 1310:
```suggestion
    "report_nft_confirm_description": "Are you sure you want to report this NFT?",
```

BrodyHughes (approved):
lgtm! the alert is nice. feels less destructive now.

magiziz (approved):

static/json/languages/en_US.json
magiziz: Tried that approach, but seems like the text break point doesn't seem too great. "Are you sure you want to report this" is in one line and then the next line you have "NFT?".

PR #1432, opened 2024-03-26
[BX-1367] Right-click accidental pass-through actions

derHowie:
Preventing radix from calling onSelect when a context click overlaps with the incoming menu.

https://www.loom.com/share/6e85c73529e54aa884bd9db6b0ea16bb

## What to test
Make sure to use the popup. I don't think this is a reproducible issue in fullscreen. Just click a bunch in an area that you know will overlap with the incoming context menu.

### DIFF
--- a/src/entries/popup/components/ContextMenu/ContextMenu.tsx
+++ b/src/entries/popup/components/ContextMenu/ContextMenu.tsx
@@ -2,7 +2,13 @@ import * as ContextMenuPrimitive from '@radix-ui/react-context-menu';
 import { DismissableLayerProps } from '@radix-ui/react-tooltip';
 import clsx from 'clsx';
 import { motion } from 'framer-motion';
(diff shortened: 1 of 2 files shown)


magiziz (approved):
Looks good. I like the 400ms timing before it can be clickable 👌

estebanmino (approved):

src/entries/popup/components/ContextMenu/ContextMenu.tsx
```diff
@@ -269,24 +278,24 @@ export const ContextMenuItem = ({
         default: 'transparent',
         hover: disabled ? 'transparent' : 'surfaceSecondary',
       }}
-      disabled={disabled}
       tabIndex={disabled ? -1 : 0}
+      disabled={disabled || isMounting}
```
greg-schrammel, line 282: I think this works
```ts
onPointerUp={(e) => e.preventDefault()}
```

PR #1435, opened 2024-03-27
Merge keychains

greg-schrammel:
Fixes BX-1602
Figma link (if any):

merge pks to it's hd group when possible
in the clip I added by pk the index 2 of a seed phrase, later when I add this seed phrase and click to create a wallet in the second index, it merges both into a single group (removes the pk keychain and merges all into the hd)

https://github.com/rainbow-me/browser-extension/assets/6232729/8433998e-5cfb-49eb-8500-9e017671532d

https://github.com/rainbow-me/browser-extension/assets/6232729/fadea7c2-c088-46da-a6fc-9a62a4a594b0

importing a duplicate pk, will just select the account

https://github.com/rainbow-me/browser-extension/assets/6232729/068ffb7e-c07a-486e-bdde-b9eec0325713

trying to import the same seed twice with prompt an error with "This seed is imported already" (maybe we need a better message here)


estebanmino (commented):
@greg-schrammel could you explain what you did here?

src/entries/popup/components/ImportWallet/ImportWalletSelection.tsx
```diff
@@ -69,17 +70,14 @@ const useDeriveAccountsFromSecrets = (secrets: string[]) => {
...
-    accountsIgnored = [],
-  }: {
-    secrets: string[];
-    accountsIgnored?: Address[];
-  }) => {
-    setIsImporting(true);
-    return (async () => {
+  const { mutateAsync: importSecrets, isLoading: isImporting } = useMutation({
+    mutationFn: async ({
+      secrets,
```
estebanmino, line 75: does this use cache to store results? if so maybe is not a good idea to use secrets for it

estebanmino (commented):
we need some qa for this

estebanmino (approved):
seems like i mentally approved this pr last week but didn't press the ✅  button

derHowie (approved):

DanielSinclair:
@greg-schrammel After you wrap up Degen Mode, if you could have a look at this and merge

PR #1444, opened 2024-03-29
fix: dont use custom_rpc param

estebanmino:
Fixes BX-####
Figma link (if any):

backend request to only pass custom_rpc for not supported networks


greg-schrammel (approved):

greg-schrammel:
why did we decide on this direction? 
even with native flashbots support I see a possible demand for custom rpcs on supported networks
like what if the user wants to use the [manifold rpc](https://www.manifoldfinance.com/) on mainnet

estebanmino:
@greg-schrammel you're right, let me do another check

estebanmino:
@greg-schrammel ok i updated it to use `getDefaultRPC` to compare

PR #1453, opened 2024-04-02
fix: deprecate isRainbowCurated for swap assets

estebanmino:
Fixes BX-####
Figma link (if any):

fixes https://rainbowhaus.slack.com/archives/C02C8JQ313N/p1712086754141749

isRainbowCurated is deprecated so i updated it to use isVerified

### DIFF
--- a/src/entries/popup/hooks/useSearchCurrencyLists.ts
+++ b/src/entries/popup/hooks/useSearchCurrencyLists.ts
@@ -283,9 +283,10 @@ export function useSearchCurrencyLists({
     ],
   );
 
+  // temporarily limiting the number of assets to display
+  // for performance after deprecating `isRainbowCurated`
   const getVerifiedAssets = useCallback(
-    (chainId: ChainId) =>
-      verifiedAssets[chainId]?.assets?.filter(({ isVerified }) => isVerified),
+    (chainId: ChainId) => verifiedAssets[chainId]?.assets?.slice(0, 50),
     [verifiedAssets],
   );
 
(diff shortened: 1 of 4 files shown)


src/entries/popup/hooks/useSearchCurrencyLists.ts
```diff
@@ -170,6 +170,8 @@ export function useSearchCurrencyLists({
       fromChainId,
     });
 
+  console.log('blastVerifiedAssets', blastVerifiedAssets);
```
greg-schrammel, line 173:
```suggestion

```

greg-schrammel (approved):

BrodyHughes (commented):
This seemed to make the asset loading on the swaps screen lag really bad. Here is the performance now
<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/5bd00f6e-41e7-4f72-a656-45e262cf9084" />

Vs Master:
<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/a51feb03-c670-4d58-8b37-8aae3b39b215" />

estebanmino:
@BrodyHughes i just added a default swap gas value that i think was causing that issue

estebanmino:
ah no is not related

DanielSinclair (approved):
There seems to be performance impact when calling `getVerifiedAssets` and filtering the whole `verifiedAssets` list (or returning that lengthy list). This list is already filtered on `isVerified` at the API request, so I instead temporarily limited the number of results to 50 which brought back performance a bit

PR #1454, opened 2024-04-02
Fix backup reminder shortcuts

BrodyHughes:
Fixes BX-1378

- this uses the `isExplainerSheet` class to determine whether or not the global shortcuts should be active
- while on an modal that `isExplainerSheet`, the `.`, `x`. `,`, etc shortcuts won't work. Only the normal navigation and `esc`. When the backup reminder is dismissed these will work again.
- this also enables the use of `esc` to close explainers

<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/ddea1c82-0f50-4393-8087-6719a23519fb" > </video>

## What to test
- `ESC` and `TAB` through items on the backup reminder
- attempt to use the `.` `,` `s` global shortcuts as well (they shouldn't work now while the backup reminder or explainer modals are up)


BrodyHughes:
@DanielSinclair want some product feedback here. There are some implications on whether we want this to impact other modals (like token expanded state) as well as the backup reminders.

DanielSinclair:
@DanielSinclair Just targeting the bottom-presenting modals like Backup Reminders should suffice. The issue was more so that shortcuts and actions were occurring invisibly behind the modal UI. We would still want the global shortcuts like `,` to work when navigating throughout layer 1 UI (where everything is presented with a left-right slide-in).

DanielSinclair:
@BrodyHughes Just targeting the bottom-presenting modals like Backup Reminders should suffice. The issue was more so that shortcuts and actions were occurring invisibly behind the modal UI. We would still want the global shortcuts like , to work when navigating throughout layer 1 UI (where everything is presented with a left-right slide-in).

src/entries/popup/hooks/useHomeShortcuts.ts
```diff
@@ -104,7 +105,9 @@ export function useHomeShortcuts() {
       const activeAppWalletSwitcher =
         appConnectionSwitchWalletsPromptIsActive();
       const inputIsFocused = getInputIsFocused();
+      const isModal = getActiveModal();
       if (inputIsFocused) return;
+      if (isModal) return;
```
DanielSinclair, line 110: Poorly named I guess, but these are actually the global shortcuts here. We'd expect `w` to open the wallet switcher from any of the layer 1 drill-downs

DanielSinclair (changes requested):

src/entries/popup/hooks/useHomeShortcuts.ts
BrodyHughes: ok cool cool. I can see what I can do in that case.

DanielSinclair (changes requested):
Seeing some strange behavior for the ESC key in other parts of the extension now. If you go to token details, press tab, then esc, you'll see a stack of explainer sheets get triggered
Screenshot 2024-04-08 at 4 57 53 PM

BrodyHughes:
> Seeing some strange behavior for the ESC key in other parts of the extension now. If you go to token details, press tab, then esc, you'll see a stack of explainer sheets get triggered Screenshot 2024-04-08 at 4 57 53 PM

ight fixed this. there was something weird going on with how we were using useReducer to change the boolean value. i just converted to useState and it works fine. useReducer is pretty unnecessary here anyways.

DanielSinclair (approved):
Looks good to me!

estebanmino (approved):

DanielSinclair:
@derHowie if you have a sec to do a code review

derHowie (approved):
💯

PR #1455, opened 2024-04-03
chore: subtracting token balance if token is hidden

magiziz:
Fixes BX-1384
Figma link (if any):

https://github.com/rainbow-me/browser-extension/assets/53529533/cd4a431b-a5cf-484b-abef-2262ab930c4c

## What to test

- Make sure token balance is subtracting when hiding them. You should see that in homepage, change wallet page and switch wallet popup (all of them are shown in video)
- Make sure to test with custom assets too

### DIFF
--- a/src/core/resources/_selectors/assets.ts
+++ b/src/core/resources/_selectors/assets.ts
@@ -104,10 +104,15 @@ export function selectUserAssetWithUniqueId(uniqueId: UniqueId) {
   };
 }
 
-export function selectUserAssetsBalance(assets: ParsedAssetsDictByChain) {
+export function selectUserAssetsBalance(
+  assets: ParsedAssetsDictByChain,
+  hidden: (asset: ParsedUserAsset) => boolean,
+) {
   const networksTotalBalance = Object.values(assets).map((assetsOnject) => {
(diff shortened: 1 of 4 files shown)


src/core/resources/_selectors/assets.ts
```diff
@@ -104,10 +104,15 @@ export function selectUserAssetWithUniqueId(uniqueId: UniqueId) {
...
+export function selectUserAssetsBalance(
+  assets: ParsedAssetsDictByChain,
+  hidden: (asset: ParsedUserAsset) => boolean,
+) {
   const networksTotalBalance = Object.values(assets).map((assetsOnject) => {
     const assetsNetwork = Object.values(assetsOnject);
+
     const networkBalance = assetsNetwork
+      .filter((asset) => !hidden(asset))
       .map((asset) => asset.native.balance.amount)
```
magiziz, line 116: You technically could use `hiddenStore.getState()` and then filter based off of the array, but that won't cause a re-render if we were to remove tokens from another component while this one stays as is.

DanielSinclair:
Functionality looks good to me!

estebanmino (approved):
lgtm!

derHowie (approved):

PR #1457, opened 2024-04-03
fix: BX-1379

estebanmino:
Fixes BX-1379
Figma link (if any):

background was trying to get remote config causing the issue described in ticket

with this pr we'll get whether proxy is enabled from popup to set up wagmi client when user is interacting with popup via background-popup messenger


derHowie (approved):

DanielSinclair (approved):
Console error disappeared for me and the popup seems to be working as expected with a little regression test. thanks!

PR #1470, opened 2024-04-05
zustand migrate util

greg-schrammel:
Fixes BX-####
Figma link (if any):

create a `migrate` util function, to apply zustand persistor migrations in order

for example if user has persisted state from **version 1** and we are now at **version 3**
this util will migrate the state from version **1** to **2** and then to **3** 
piping each resulted state into the next migration function

the first function arg must be typed, the rest is deriving it's arg value from the previous function return 
so the first persisted state is not type safe but from the second migration on it should always be, 
the zustand types will make sure the last migration function returns the right type for the most recent store state

we can use also use named functions if it makes it easier to understand
```ts
...
version: 3
migrate: migrate(
  // version 1 didn't need a migration
  function v1(state: FavoritesState) {
    return state;
  },
  // version 2 added avalanche
  function v2(state) {
    return mergeNewOfficiallySupportedChainsState(state, [
      ChainId.avalanche,
    ]);
  },
  // version 3 added blast
  function v3(state) {
    return mergeNewOfficiallySupportedChainsState(state, [ChainId.blast]);
  },
),
...
```


DanielSinclair (approved):
Nice!

estebanmino (approved):
seems to be working fine, maybe we need some qa

DanielSinclair:
@greg-schrammel Can you revisit for the Degenchain merge conflict? I think we may need to stage this in front of Esteban's fix

PR #1471, opened 2024-04-08
hide simulation details tab when thers nothing to show in it

greg-schrammel:
Fixes BX-####
Figma link (if any):

hide simulation details tab when there is nothing to show in it


estebanmino:
can we hide it in case there's no simulation details?


https://github.com/rainbow-me/browser-extension/assets/12115171/729a72fa-fc64-4e93-9a5e-c51c2c791929

imo feels wrong that we start with two tabs and go into three tabs for most of the requests

greg-schrammel:
> can we hide it in case there's no simulation details?
> 
>  Screen.Recording.2024-04-09.at.14.14.54.mov 
> imo feels wrong that we start with two tabs and go into three tabs for most of the requests

oh shit didn't mean to do that, good catch

magiziz (approved):
Same suggestion from Esteban, but apart from that lgt 🙏

DanielSinclair (approved):

PR #1479, opened 2024-04-09
fix: handle provider request isSupportedChain

estebanmino:
Fixes BX-####
Figma link (if any):

isSupportedChain was only checking default chains so at the moment of switch chain, if the new chain was a custom chain you wouldn't be able to connect

before 

https://github.com/rainbow-me/browser-extension/assets/12115171/c65e5a2e-a217-4c14-9d93-6075074dcd74

now

https://github.com/rainbow-me/browser-extension/assets/12115171/85dc145c-10b4-4650-a0fe-892a98122d50


derHowie (approved):

greg-schrammel (approved):
maybe we should rename `isSupportedChain` later, or keep the name and add logic to check if theres an rpc configured for it, or name it something referencing that it has first class support

PR #1486, opened 2024-04-10
fix tx request native gas balance of custom networks

greg-schrammel:
Fixes BX-####
Figma link (if any):

makes `useCustomNetworkAsset` accept a address arg so we can use the app session address instead of the `currentAddress` (that's the one selected in the wallet, but not necessarily the connected to the request dapp)

this was making so that sometimes a request from a "custom network" displayed "buy gas token" instead of the simulation and blocked the user from signing


estebanmino (approved):
wow nice 💯

derHowie (approved):
👍

magiziz (approved):
Awesome 🤩

PR #1494, opened 2024-04-13
ledger cannot read properties of null "decorateAppAPIMethods"

greg-schrammel:
Fixes BX-####
Figma link (if any):

this is a hard bug to debug as it doesn't always happen
but I tracked it down to believe it's something to do with a stale clojure in this request listener
could be completely wrong tho 🤷‍♂️🙃

## What to test

try to swap using a ledger


BrodyHughes (approved):
Tested:
- Onboarding
- L1 swap + send
- L2 swap + send

PR #1496, opened 2024-04-15
fix: add zora again if is missing

estebanmino:
Fixes BX-1420
Figma link (if any):

adding zora back to the state if is missing for these users

details and POW  video in ticket

### DIFF
--- a/src/core/state/rainbowChains/index.ts
+++ b/src/core/state/rainbowChains/index.ts
@@ -1,4 +1,4 @@
-import { Chain } from 'viem/chains';
+import { Chain, zora } from 'viem/chains';
 import create from 'zustand';
 
 import { SUPPORTED_CHAINS, getDefaultRPC } from '~/core/references';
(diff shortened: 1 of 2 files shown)


greg-schrammel (commented):
this has the same issue I tried to solve with this https://github.com/rainbow-me/browser-extension/pull/1470

if the last time the user opened the bx this store was at version 2, the only migration that's happening is the one adding support for blast

I think we can fix this by using 
```diff
- if (version === 4)
+ if (version <= 4)
```
or merging that util

estebanmino:
@greg-schrammel i'm down to merfge that, is already approved

DanielSinclair (approved):
Tested from the previous release (with existing Zora RPC) and worked for that

BrodyHughes (approved):
got my app into a broken state (Zora RPC gone) and this fixed it! nice work.

<video src="https://github.com/rainbow-me/browser-extension/assets/41711440/a998f58e-8bea-4ec7-a00c-7ae5446a3d12"> </video>

DanielSinclair:
@estebanmino Added to the merge queue after our testing FYII

PR #1498, opened 2024-04-15
[BX-1321] feat: show warning when asset price unknown

derHowie:
Added a warning whenever we do not receive a price for an asset being swapped. I based the copy and design on the app's work [here](https://linear.app/rainbow/issue/APP-1295/update-slippage-warning-we-show-when-swapping-tokens-without-price).

Screen Shot 2024-04-15 at 7 22 06 PM

## What to test
Ensure that the warning appears when setting up a swap including an asset with no reported price.

### DIFF
--- a/src/entries/popup/pages/swap/index.tsx
+++ b/src/entries/popup/pages/swap/index.tsx
@@ -161,6 +161,58 @@ const SwapWarning = ({
   );
 };
 
+const MissingPriceExplanation = ({
+  assetToBuy,
+  assetToSell,
+}: {
+  assetToBuy: ParsedSearchAsset | null;
+  assetToSell: ParsedSearchAsset | null;
+}) => {
+  const missingPriceSymbols = [assetToBuy, assetToSell].reduce(
+    (symbols, asset) => {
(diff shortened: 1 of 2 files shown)


estebanmino (approved):
lgtm

src/entries/popup/pages/swap/index.tsx
```diff
@@ -161,6 +161,58 @@ const SwapWarning = ({
...
+  const missingPriceSymbols = [assetToBuy, assetToSell].reduce(
+    (symbols, asset) => {
+      if (asset?.price?.value === undefined && asset?.symbol) {
+        return [...symbols, asset?.symbol];
+      }
+      return symbols;
+    },
+    [] as string[],
+  );
+  if (!missingPriceSymbols?.length) return null;
```
greg-schrammel, line 180: I don't think this can ever be undefined?
```suggestion
  if (!missingPriceSymbols.length) return null;
```

greg-schrammel (approved):

magiziz (approved):
Nice 🙏

PR #1499, opened 2024-04-16
combined assets by uniqueId

estebanmino:
Fixes BX-1421
Figma link (if any):

details of bug in ticket

maybe this is still happening since we're combining assets in an array directly from assets and customNetworkAssets, with this pr we'll only get one per uniqueId so there won't be duplicates for sure

image

### DIFF
--- a/src/entries/popup/hooks/send/useSendAsset.ts
+++ b/src/entries/popup/hooks/send/useSendAsset.ts
@@ -1,4 +1,3 @@
-import uniqBy from 'lodash/uniqBy';
 import { useCallback, useMemo, useState } from 'react';
 
 import {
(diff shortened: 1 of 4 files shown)


src/entries/popup/hooks/send/useSendAsset.ts
```diff
@@ -65,14 +64,27 @@ export const useSendAsset = () => {
...
     () =>
-      uniqBy([...assets, ...customNetworkAssets], 'uniqueId').sort(
+      combinedAssets.sort(
         (a: ParsedUserAsset, b: ParsedUserAsset) =>
           parseFloat(b?.native?.balance?.amount) -
           parseFloat(a?.native?.balance?.amount),
       ),
-    [assets, customNetworkAssets],
+    [combinedAssets],
   );
```
magiziz, line 88: Do you know if this would impact hidden and pin functionality ? Since if there are two same `uniqueId`'s and i hide one of them both of them will be hidden since we hide by `uniqueId`. I might be wrong, but can you test that ?

magiziz (changes requested):
Just one small question. Otherwise everything looks good to me 👍

src/entries/popup/hooks/send/useSendAsset.ts
estebanmino: that would be the expected right? since we hide uniqueid

estebanmino: Uploading Screen Recording 2024-04-17 at 14.14.49.mov…

how do i uhide? lol

estebanmino: https://github.com/rainbow-me/browser-extension/assets/12115171/90aa88ed-c84a-4993-9f51-1f6fe32b9eda

magiziz: Ahh i see. Alright then it looks great to me 🙏

derHowie (approved):
👌

magiziz (approved):

src/entries/popup/hooks/send/useSendAsset.ts
magiziz: Opps. Approved 🙏

PR #1500, opened 2024-04-16
pass currency in tx simulation

estebanmino:
Fixes BX-1320
Figma link (if any):

details in ticket

passing currency in tx simultaion request

image


greg-schrammel (approved):
do we have this feat for txs too? this is adding for singnatures only right? would be cool to add to tx requests

magiziz (approved):
lgtm

PR #1507, opened 2024-04-18
[BX-1423] fix: set max width on app connection wallet switcher names

derHowie:
Added a max width to wallet name dropdown to prevent 2 lines

BEFORE:
Screen Shot 2024-04-18 at 10 40 58 AM
AFTER:
Screen Shot 2024-04-18 at 10 36 36 AM

## What to test
Make a really long wallet name, then make sure the dropdown still displays within a single line


src/entries/popup/pages/messages/BottomActions/index.tsx
```diff
@@ -46,8 +46,14 @@ export const WalletName = ({
 }: { address: Address } & Partial<TextProps>) => {
   const { displayName } = useWalletInfo({ address });
   return (
-    // eslint-disable-next-line react/jsx-props-no-spreading
-    <TextOverflow color="label" size="14pt" weight="semibold" {...props}>
+    <TextOverflow
+      color="label"
+      size="14pt"
+      weight="semibold"
+      maxWidth={130}
```
estebanmino, line 53: the original idea for TextOverflow was to not set a `maxWidth`, any other idea to fix the TextOverflow component so this doesn't happen again in other place?

mikedemarais: @estebanmino can you explain further? 

TextOverflow seems to accept maxWidth prop https://github.com/rainbow-me/browser-extension/blob/150dfaa18d622d67140f5aa770fa3c1d2a7dc040/src/design-system/components/TextOverflow/TextOverflow.tsx#L35 

but also im not necessarily suggesting that we continue to use maxWidth here,, typically these types of truncated text components rely on flexbox to achieve desired effect. 

@derHowie if you add flex: 1 to this component will that fix the problem?

derHowie: > the original idea for TextOverflow was to not set a maxWidth, any other idea to fix the TextOverflow component so this doesn't happen again in other place?

@estebanmino, I take your point about `maxWidth` since it can be brittle. I tried my hand at altering `TextOverflow` directly, but given all the various containers, siblings, and display modes the component could find itself in, I'm unsure we can add a combination of styles that will handle every scenario. For example, adding @mikedemarais's suggestion works for the wallet name in this ticket, but breaks things in several other cases.

I think that the approach should be to use a container that scales dynamically whenever we use TextOverflow. This creates the effect we're going for, but isn't as reusable out of the box as what you're describing.

greg-schrammel: have you tried adding `wrap={false}` to the wrapping `<Inline />`?

estebanmino: @greg-schrammel 's solution works for me 
image

derHowie: okay @estebanmino @greg-schrammel, I went with your simplified version

estebanmino (approved):

greg-schrammel (approved):

PR #1509, opened 2024-04-22
remove white flash screen on startup

estebanmino:
Fixes BX-####
Figma link (if any):

removing flash white screen on startup

with this pr we'll be initializing theme from index.html

before we were doing this from `initThemingCritical` and `initThemingBody` causing the flash since those take a fraction of a second after dom is loaded to be initialized

https://www.loom.com/share/f0e221cff41b4f188cc00462cad61fd2

### DIFF
--- a/src/design-system/styles/initThemingLocal.ts
+++ b/src/design-system/styles/initThemingLocal.ts
@@ -1,0 +1,9 @@
+import { currentThemeStore } from '~/core/state/currentSettings/currentTheme';
+
+export function initThemingLocal() {
+  const localTheme = localStorage.getItem('theme');
+  if (!localTheme) {
+    const currentTheme = currentThemeStore.getState().currentTheme;
+    localStorage.setItem('theme', currentTheme);
+  }
+}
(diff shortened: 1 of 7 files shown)


src/design-system/styles/initThemingLocal.ts
```diff
@@ -0,0 +1,9 @@
+import { currentThemeStore } from '~/core/state/currentSettings/currentTheme';
+
+export function initThemingLocal() {
+  const localTheme = localStorage.getItem('theme');
+  if (!localTheme) {
+    const currentTheme = currentThemeStore.getState().currentTheme;
```
greg-schrammel, line 6: is the extra localStorage set/get needed? can't just use the zustand store like this?

derHowie: guessing this is so the static vanilla js scripts have somewhere to read from

derHowie (approved):
🙏

src/design-system/styles/initThemingLocal.ts
estebanmino: yeah exactly what @derHowie says

greg-schrammel: aah zustand is using an async `chrome.storage.local` for everything, maybe we could consider opting for standard `localStorage` when fit
so the store is already hydrated when created vs having to wait for the async run to hydrate it

PR #1512, opened 2024-04-23
fix degen native asset

estebanmino:
Fixes BX-1428
Figma link (if any):

we were picking ETH as native asset for degen making prices go crazy

since degen doesn't exist in mainnet i had to refactor the way we fetch native assets, so we now `requestMetadata` of the native asset address on the chain together with the chainid, before we were just picking the assets for mainnet making it impossible to fetch native assets that don't exist in mainnet like degen

plz make sure all chains work fine, attached a video of my qa

https://www.loom.com/share/b6328a49d83b4d29b50f79d8e7ff986d

### DIFF
--- a/src/core/references/index.ts
+++ b/src/core/references/index.ts
@@ -151,7 +151,7 @@ export const WETH_BLAST_ADDRESS = '0x4300000000000000000000000000000000000004';
 export const USDB_BLAST_ADDRESS = '0x4300000000000000000000000000000000000003';
 
 // degen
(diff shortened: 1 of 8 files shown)


derHowie (approved):
lgtm 👍

DanielSinclair (approved):
Did a test swap between DEGEN and GOAT 🐐  lgtm

DanielSinclair:
Polygon and Optimism swaps also look good

PR #1519, opened 2024-04-25
Activity details loading

greg-schrammel:
Fixes BX-####
Figma link (if any):

- activity details sheet insta shows on clicking an activity (before was waiting for loading), now it shows with the data we got and updates as more data comes
- removed `testnetMode` from the consolidatedTransactionsQueryKey as it was not used in the query itself and I didn't find a reason to have it in the query key lmk if Im mistaken
- refactored a lil bit the `useTransaction` hook, aggregated into a single useQuery for either supported or unsupported BE networks and added some more info we could find when fetching the tx data from the provider

### DIFF
--- a/src/core/resources/transactions/transaction.ts
+++ b/src/core/resources/transactions/transaction.ts
@@ -1,6 +1,6 @@
 import { Provider, TransactionResponse } from '@ethersproject/providers';
 import { formatUnits } from '@ethersproject/units';
-import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { QueryClient, useQuery, useQueryClient } from '@tanstack/react-query';
 import { Hash, getProvider } from '@wagmi/core';
 import { Address } from 'wagmi';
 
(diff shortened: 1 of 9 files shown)


derHowie (approved):

src/core/resources/transactions/transaction.ts
```diff
@@ -52,6 +51,14 @@ export const fetchTransaction = async ({
   currency: SupportedCurrencyKey;
   chainId: ChainId;
 }) => {
+  if (!SUPPORTED_CHAIN_IDS.includes(chainId)) {
```
estebanmino, line 54: did you find any case where we pass unsopperted chainids?

all of the instances are using chains coming from useBackendSupportedChains hook like this https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/hooks/useTransactionListForPendingTxs.ts#L32 or from that constant https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/pages/home/Approvals/Approvals.tsx#L342

src/core/resources/transactions/transaction.ts
```diff
@@ -83,133 +94,93 @@ export const fetchTransaction = async ({
...
+
+      gasUsed: receipt.gasUsed.toString(),
+
+      feeType: !block.baseFeePerGas ? 'legacy' : 'eip-1559',
+      fee: receipt.cumulativeGasUsed.mul(receipt.effectiveGasPrice).toString(),
+      baseFee: block.baseFeePerGas?.toString(),
+    };
+  }
+
+  const type = await guessTransactionType(provider, transaction);
```
estebanmino, line 173: why await?

src/core/resources/transactions/transaction.ts
```diff
@@ -218,32 +189,53 @@ export const useTransaction = ({
...
+        if (tx) return tx;
+      }
+
+      // try to find in custom network transactions store
+      const { getCustomNetworkTransactions } =
+        customNetworkTransactionsStore.getState();
+      return getCustomNetworkTransactions({ address }).find(
+        (tx) => tx.hash === hash && tx.chainId === chainId,
+      );
+    },
```
estebanmino, line 234: i don't follow why this is mixing txs supported by backend and the ones from `customNetworkTransactionsStore`, there are separate stores and we're using them separately

src/core/resources/transactions/transaction.ts
greg-schrammel: I removed the `useBackendSupportedChains` hook didn't understand why we had two hooks to fetch txs, replaced with a single `useTransaction` (which from outside this file works the same) to handle the transaction metadata no matter where that data is coming from

src/core/resources/transactions/transaction.ts
greg-schrammel: Im fetching data from the provider to guess the type
like `transaction.to` has code (`provider.getCode`) means it's probably a transaction interacting with a contract

greg-schrammel: hmm idk whats the expected behavior on smart wallets

src/core/resources/transactions/transaction.ts
estebanmino: okok it makes sense

estebanmino (approved):
lgtm 🚀

PR #1520, opened 2024-04-25
update migrations

estebanmino:
Fixes BX-1443
Figma link (if any):

new method `migrate` is working fine but we were using it wrongly together with version number for state slices

for rainbowChains we had version 7, with 6 migrations

`migrate` applies the migration functions one by one removing the ones from 0 to `version` (.toSliced(0,version)`) so if you were in version 6, all of the migrations would be removed since the migrations were 6, so none of the migrations were being executed 

we have to have this in mind for the next time we use this method

I updated `migrate` to accept any arbitrary number of elements in an array param since it was annoying to add a new interface for each element in the array

### DIFF
--- a/src/core/utils/migrate.ts
+++ b/src/core/utils/migrate.ts
@@ -16,7 +16,7 @@ interface Migrator {
 export const migrate: Migrator = (migrations: ((s: any) => any)[]) => {
   return (persistedState: any, version: number) => {
     return migrations
-      .slice(0, version) // Use slice to select up to the specified version
+      .toSpliced(0, version)
       .reduce((acc, fn) => fn(acc), persistedState);
   };
 };
--- a/src/core/utils/migrate.ts
+++ b/src/core/utils/migrate.ts
@@ -3,31 +3,7 @@
 type R<T> = (persistedState: unknown, version: number) => T;
 
 interface Migrator {
-  <A>(m1: (s: any) => A): R<A>;
-  <A, B>(m1: (s: any) => A, m2: (s: A) => B): R<B>;
-  <A, B, C>(m1: (s: any) => A, m2: (s: A) => B, m3: (s: B) => C): R<C>;
-  <A, B, C, D>(
-    m1: (s: any) => A,
-    m2: (s: A) => B,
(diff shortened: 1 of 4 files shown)


derHowie (approved):

src/core/utils/userChains.ts
```diff
@@ -80,7 +80,7 @@ export const chainLabelMap: Record<
   [ChainId.zora]: [ChainNameDisplay[zoraSepolia.id]],
   [ChainId.avalanche]: [ChainNameDisplay[avalancheFuji.id]],
   [ChainId.blast]: [ChainNameDisplay[chainBlastSepolia.id]],
-  [ChainId.degen]: [ChainNameDisplay[chainDegen.id]],
+  [ChainId.degen]: [],
```
greg-schrammel, line 83: should this be removed?

greg-schrammel:
> so if you were in version 6, all of the migrations would be removed since the migrations were 6, so none of the migrations were being executed

yes this is the expected behavior

we can make the util throw if the version and quantity of migrations mismatches

greg-schrammel:
tried an idea here https://github.com/rainbow-me/browser-extension/pull/1523

src/core/utils/userChains.ts
estebanmino: yeah

src/core/utils/migrate.ts
```diff
@@ -3,31 +3,7 @@
...
-  <A, B, C, D, E, F>(
-    m1: (s: any) => A,
-    m2: (s: A) => B,
-    m3: (s: B) => C,
-    m4: (s: C) => D,
-    m5: (s: D) => E,
-    m6: (s: E) => F,
-  ): R<F>;
-  // if you need more migrations, add more overloads here
+  <T>(migrations: ((s: any) => any)[]): R<T>;
```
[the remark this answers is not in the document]
estebanmino: is there any other option for this? adding 15+ lines per new migration is not the best either

estebanmino:
closing for https://github.com/rainbow-me/browser-extension/pull/1523

PR #1523, opened 2024-04-26
refactor migrate util into persistOptions util

greg-schrammel:
Fixes BX-####
Figma link (if any):

replaces the util `migrate` with a `persistOptions` so we have more control over the persist options and can make sure it matches the current version with the number of migrations

this could probably be integrated into `createStore` but I couldn't make the types work right rn


estebanmino (approved):

estebanmino:
we also need this change https://github.com/rainbow-me/browser-extension/pull/1520#discussion_r1581552147 to remove the "degen" label here 
image

derHowie (approved):
cool idea 👌

PR #1526, opened 2024-04-29
feat: contacts

magiziz:
Fixes BX-1429
Figma link (if any):

https://github.com/rainbow-me/browser-extension/assets/53529533/c1622099-ce92-4b88-a459-614939574509

## What to test

- Make sure you can add contacts for any current wallets in Cmd+K menu for `My Wallets` tab
- When you add a contact make sure you can't add the same contact twice in `My Wallets` tab
- You should see your recent added contacts in `My Contacts` section in Cmd+K menu
- If you were to click on a contact or press Enter you should be redirected to `/send` page and the from address should be the contact address

image

- Test out different contact actions for a contact (Watch wallet, send, remove contact, copy and view on etherscan)

image

- **Bonus.** Currently there is a bug in production where if you add a contact without a name it shows you a blank state, in this PR it should fix that ✅

image


src/entries/popup/components/CommandK/useSearchableContacts.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const allowSend = useMemo(
+    () => !isWatchingWallet || featureFlags.full_watching_wallets,
+    [featureFlags.full_watching_wallets, isWatchingWallet],
+  );
+
+  const shouldNavigateToSend = useMemo(() => {
+    // Trezor should always be in a new tab
+    return !(isTrezor && !isFullScreen) && allowSend;
+  }, [allowSend, isFullScreen, isTrezor]);
```
greg-schrammel, line 53: is there a reason to memoize these primitive values?

src/core/state/contacts/index.ts
```diff
@@ -10,6 +10,7 @@ export interface Contact {
 
 export interface ContactsStore {
   contacts: { [address: Address]: Contact };
+  selectedContactAddress: Address | null;
```
greg-schrammel, line 13: instead of holding this in a store what do you think of using a search param in the page? 
like navigating to `${ROUTES.send}?to=${contactAddress}`

src/entries/popup/components/CommandK/useSearchableContacts.ts
magiziz: Yeah i just copied from [here](https://github.com/rainbow-me/browser-extension/blob/1f33946170e69e71344a18246da3ab6f9f95455e/src/entries/popup/pages/home/Header.tsx#L160-L163)

src/core/state/contacts/index.ts
magiziz: Yeah that'd be a better approach, but since i saw we use states all over the place i choose to stick with that approach to not complicate things with states and queries. [Here](https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/pages/send/index.tsx#L490-L501) is an example.

nickbytes: If there's a better approach, do the better approach @magiziz. We do not need to copy things from other portions of the application if there's a better way.

src/entries/popup/components/CommandK/useSearchableContacts.ts
nickbytes: @magiziz outside of copying code from another area, is there a reason to memoize these values?

src/core/state/contacts/index.ts
magiziz: Yep. Queries should be much easier in this. Changes are pushed 🙏

src/entries/popup/components/CommandK/useSearchableContacts.ts
magiziz: Nope i don't think we need them to be memoized since it's just returning a boolean and booleans shouldn't be memoized. Pushed a fix for this 👍

src/entries/popup/pages/send/ToAddressInput.tsx
```diff
@@ -106,10 +106,14 @@ const WalletRow = ({
...
-  });
-  const name = section === 'contacts' ? contactName : displayName;
+  const { displayName, contactAddress, contactName, isNameDefined } =
+    useWalletInfo({
+      address: wallet,
+    });
+  const name =
+    section === 'contacts'
+      ? contactName || truncateAddress(contactAddress)
+      : displayName;
```
magiziz, line 116: This just fixes the bug we have in our "Contacts" menu on send page

src/entries/popup/pages/send/index.tsx
```diff
@@ -111,6 +112,9 @@ export function Send() {
   const { isContact } = useContactsStore();
   const { allWallets } = useWallets();
   const { hiddenAssets } = useHiddenAssetStore();
+  const [urlSearchParams] = useSearchParams();
+
+  const queryToAddress = urlSearchParams.get('to');
```
greg-schrammel, line 117: I think that since this can technically be edited by the user we should check if it's a valid address/ens

magiziz: Good catch 🙏

greg-schrammel (approved):

estebanmino (approved):
code lgtm, i tried to break it but seems to be working as expected 💯

DanielSinclair:
@magiziz There was an odd race condition when the merge queue had merged this that broke CI. Pushed up a CI fix here: #1539

PR #1531, opened 2024-05-02
feat: upgrading swap sdk to 0.19.0

fringlesinthestreet:

Upgraded sdk version to 0.19.0, now supporting reservoir relay instant bridges

Supported networks:
* socket
    * mainnet
    * OP
    * BSC
    * Arbitrum
    * Polygon
    * Base
    * Avalanche
* relay
    *  mainnet: eth, usdc
    * OP: eth, usdc
    * Arbitrum: eth, usdc
    * Base: eth, usdc, degen
    * Blast: eth
    * Zora: eth
    * Degen: degen

## What to test

Swaps, crosschain swap and bridges flows
* relay supports bridging eth <-> eth, usdc <-> usdc and degen <-> degen
    * For relay's ERC20 bridging, no approval tx is needed

## Notes

Need to publish new sdk version on [npm](https://www.npmjs.com/package/@rainbow-me/swaps)


src/core/raps/unlockAndCrosschainSwap.ts
```diff
@@ -51,7 +51,8 @@ export const estimateUnlockAndCrosschainSwap = async (
     isLowerCaseMatch(ETH_ADDRESS_AGGREGATOR, sellTokenAddress) ||
     isNativeAsset(assetToSell.address, chainId);
 
-  if (!isNativeAssetUnwrapping && !nativeAsset) {
+  const shouldNotUnlockAsset = quote.no_approval !== undefined && quote.no_approval;
```
greg-schrammel, line 54: I get confused with double negatives `!shouldNotUnlockAsset` is the same as `shouldUnlockAsset`?
no_approval === true means it does not have an approval and need one
no_approval === false means we good to go
no_approval === undefined means what

fringlesinthestreet: Srry for the double negative. @greg-schrammel I'll try to explain a bit:

`shouldNotUnlockAsset` values:
- `true`: marks that the asset does not need an approval/permit
- `false`: **does NOT mean that the asset requires unlock**. Instead, it means that we **need to check** if the asset requires unlock.

------

- `no_approval === true`: means the asset never needs an approval/permit
- `no_approval === false`: means the asset may need an approval/permit, should be checked
    - `no_approval === undefined`: means the backend did not return this field, so should be treated as it was false.

greg-schrammel: can we change it to `const shouldUnlockAsset = !quote.no_approval;`?

greg-schrammel (approved):

derHowie (approved):

estebanmino (commented):
i believe we need these same changes in https://github.com/rainbow-me/browser-extension/blob/master/src/core/raps/unlockAndCrosschainSwap.ts#L55C1-L56C1 otherwise bridges won't work as expected

estebanmino:
also here https://github.com/rainbow-me/browser-extension/blob/master/src/core/raps/unlockAndSwap.ts#L151 and https://github.com/rainbow-me/browser-extension/blob/master/src/core/raps/unlockAndSwap.ts#L151 otherwise we'll be sending the approve tx anyways

DanielSinclair:
@estebanmino Looks like a small type issue popped up

PR #1532, opened 2024-05-03
probably fix overlapping activities

greg-schrammel:
Fixes BX-####
Figma link (if any):

pretty sure the bug was in the way we were using framer motion `layoutId` to transition between lists, the idea was set `layoutId` to the list row index so when we change tabs between activities and tokens, the same index rows would transition smoothly
the issue is when we are adding or reordering stuff in each list, like tx A is index 1 in the list, a new pending tx B is added, now B is index 1, so framer motion is trying to animate tx row A to B but it shouldn't!! 
that's the reason we don't use index as the `key` in react, we shouldn't use as `layoutId` either

to replace the previous transition I added this one

https://github.com/rainbow-me/browser-extension/assets/6232729/670ca442-776b-4228-a48f-d29e4e7b01d3

### DIFF
--- a/src/entries/popup/pages/home/Activity/ActivitiesList.tsx
+++ b/src/entries/popup/pages/home/Activity/ActivitiesList.tsx
@@ -131,9 +131,9 @@ export function Activities() {
       >
         <Box
           width="full"
+          position="relative"
           style={{
             height: activityRowVirtualizer.getTotalSize(),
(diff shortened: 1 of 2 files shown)


estebanmino (approved):
👌

magiziz (approved):
✨

PR #1535, opened 2024-05-06
Home rerenders

greg-schrammel:
Fixes BX-####
Figma link (if any):

optimizes a bit the rerenders of the Home route

I think a nice pattern we can adopt I started doing in this pr is to do components instead of hooks for stuff like watchers/listeners that don't directly update the state of the component we are calling that hook in

example created this component
```ts
export function SwitchWalletShortcuts() {
  useSwitchWalletShortcuts();
  return null;
}
```
before we were calling `useSwitchWalletShortcuts()` in the Home component
now we "render" `<SwitchWalletShortcuts />`

this encapsulates the updates to this single component, instead of the whole `<Home />` component having to rerender because of a single leaky update in that hook, just `<SwitchWalletShortcuts />` rerenders


src/entries/popup/hooks/useHomeShortcuts.ts
```diff
@@ -246,3 +246,8 @@ export function useHomeShortcuts() {
     handler: handleHomeShortcuts,
   });
 }
+
+export function HomeShortcuts() {
+  useHomeShortcuts();
+  return null;
+}
```
estebanmino, line 253: nice, i saw this in your tx animation pr and was gonna ask about it 
👌

magiziz: This is really nice and creative! Now we can have `<HomeShortcuts />` re-rendering and not whole `<Home />` component 🚀

magiziz (approved):
Nice Greg! I'll start using the same pattern when adding new hooks 👍

derHowie (approved):

christianbaroni:
Good stuff, I'm a fan of that pattern

PR #1537, opened 2024-05-06
fix: Cmd+K not working with latest i18n strings

magiziz:
Fixes BX-1112
Figma link (if any):

1. The issue occurs when we use the `i18n` singleton with `useMemo`, which memoizes the value and doesn't get the latest i18n strings for the updated language. 

2. Another problem is we keep object references outside the components. Meaning when a component is re-rendered, it doesn't get the latest object values from that same object reference. Specifically the objects causing this issue are `PAGES` and `actionLabels`.

**Before**

before

**After**

after

## What to test

- Go to Settings -> Language -> _pick any language_ -> do Cmd+K and you should see the translated words from the language you've picked


src/entries/popup/components/CommandK/pageConfig.ts
```diff
@@ -3,69 +3,69 @@ import { i18n } from '~/core/languages';
...
+  listTitle: (command: SearchItem | null) => string;
+  searchPlaceholder: () => string;
+  emptyLabel?: () => string;
 }
 
 export const PAGES: { [KEY: string]: Page } = {
   HOME: {
-    listTitle: i18n.t('command_k.pages.home.section_title'),
-    searchPlaceholder: i18n.t('command_k.pages.home.search_placeholder'),
+    listTitle: () => i18n.t('command_k.pages.home.section_title'),
```
magiziz, line 13: Same rule applies here. If same object reference is used outside the component and re-render occurs we'll not get the latest `i18n.t(...)` string value. We'll use function as a method here to forcefully call it during the component re-render, so we can get the latest i18n string.

src/entries/popup/components/CommandK/CommandKToolbar.tsx
```diff
@@ -105,7 +105,7 @@ export function CommandKToolbar({
                   whileTap={{ scale: 0.925 }}
                 >
                   <Inline alignVertical="center" space="10px" wrap={false}>
-                    <Inline alignVertical="center" space="4px">
+                    <Inline wrap={false} alignVertical="center" space="4px">
```
magiziz, line 108: This change solved this bug i had

Screenshot 2024-05-06 at 15 17 35

greg-schrammel:
shit this made me realize we shouldn't add `title` and `description` to the `RainbowTransaction` type
`parseTransaction` sets it in the queryFn which is then persisted so the translations wont update until the next refetch
it should be derived on use, I think it's an easy fix if you wanna take it

magiziz:
> shit this made me realize we shouldn't add `title` and `description` to the `RainbowTransaction` type `parseTransaction` sets it in the queryFn which is then persisted so the translations wont update until the next refetch it should be derived on use, I think it's an easy fix if you wanna take it

Ah i see. Yeah i'll include it in this PR as well.

estebanmino (approved):
lgtm 👌

greg-schrammel (approved):

PR #1539, opened 2024-05-09
fix: open profile command for contacts cmdk

DanielSinclair:
Fixes BX-####
Figma link (if any):

- fixed a ci/linting issue introduced on #1526 related to command k actions
- there is apparently a race condition for a few seconds/mins with the merge queue when pushing a new commit and ci checks from a previous commit finishing up, where untested commits can get merged

### DIFF
--- a/src/entries/popup/hooks/useInfiniteTransactionList.ts
+++ b/src/entries/popup/hooks/useInfiniteTransactionList.ts
@@ -33,7 +33,7 @@ export const useInfiniteTransactionList = ({
   const { currentAddress: address } = useCurrentAddressStore();
   const { currentCurrency: currency } = useCurrentCurrencyStore();
   const pendingTransactions = usePendingTransactionsStore(
-    (s) => s.pendingTransactions[address],
+    (s) => s.pendingTransactions[address] ?? [],
   );
   const [manuallyRefetching, setManuallyRefetching] = useState(false);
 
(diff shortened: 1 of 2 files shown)


magiziz (approved):
lgtm

greg-schrammel (approved):

src/entries/popup/hooks/useInfiniteTransactionList.ts
```diff
@@ -33,7 +33,7 @@ export const useInfiniteTransactionList = ({
   const { currentAddress: address } = useCurrentAddressStore();
   const { currentCurrency: currency } = useCurrentCurrencyStore();
   const pendingTransactions = usePendingTransactionsStore(
-    (s) => s.pendingTransactions[address],
+    (s) => s.pendingTransactions[address] ?? [],
```
greg-schrammel, line 36: I thought about this when doing this change, thought it was fine to leave as undefined

since it's not, and I'm not sure if I think it's a beautiful solution yet, instead of recreating the array on every selector run, we could return a stable empty array 

```ts
const stableEmptyPendingTransactionsArray = []
export const useInfiniteTransactionList = ({ ... }) =>
  ...
  const pendingTransactions = usePendingTransactionsStore(
      (s) => s.pendingTransactions[address] || stableEmptyPendingTransactionsArray,
   );
```
tbh I dont think I like how this reads, the way I'd have to jump array to look what this `stableEmptyPendingTransactionsArray` means/is, but as of rn is the best way I think to not need unnecessary rerenders

this way we can use `pendingTransactions` in the dependency array at `const formattedTransactions = useMemo(` without rerunning this memo every render

greg-schrammel (approved):

PR #1542, opened 2024-05-14
fix: `useKeyboardShortcut` not working with Cmd+K backspace shortcut

magiziz:
Fixes BX-1465

If you try to go to different menus in Cmd+K and then click backspace shortcut you won't get navigated back to where you was. There is also a bug where you can't copy NFT token id using `c` shortcut.

## What to test

Open Cmd+K menu -> navigate to "My Wallets" -> hit backspace shortcut and you should be able to get navigated back to the default Cmd+K view

### DIFF
--- a/e2e/parallel/shortcuts-home.test.ts
+++ b/e2e/parallel/shortcuts-home.test.ts
@@ -216,6 +216,27 @@ describe.runIf(browser !== 'firefox')(
       expect(txHash).toBe(false);
     });
 
+    it('should navigate to send page using Cmd+K', async () => {
+      await executePerformShortcut({ driver, key: 'k' });
+      await executePerformShortcut({ driver, key: 'ENTER' });
+      await checkExtensionURL(driver, 'send');
+    });
+
+    it('should navigate to my wallets section in Cmd+K', async () => {
+      await goToPopup(driver, rootURL);
+      await executePerformShortcut({ driver, key: 'k' });
+      await executePerformShortcut({
+        driver,
+        key: 'ARROW_DOWN',
+        timesToPress: 2,
+      });
+      await executePerformShortcut({ driver, key: 'ENTER' });
+      await findElementByText(driver, 'Switch to Wallet');
+      await executePerformShortcut({ driver, key: 'BACK_SPACE' });
+      await findElementByText(driver, 'Suggestions');
+      await executePerformShortcut({ driver, key: 'ESCAPE' });
+    });
+
     it('should be able to lock extension with keyboard', async () => {
       await executePerformShortcut({ driver, key: 'DECIMAL' });
       await executePerformShortcut({ driver, key: 'l' });
--- a/src/entries/popup/hooks/useKeyboardShortcut.ts
+++ b/src/entries/popup/hooks/useKeyboardShortcut.ts
@@ -1,4 +1,4 @@
-import { useEffect, useLayoutEffect, useMemo, useRef } from 'react';
+import { useEffect, useMemo } from 'react';
 
 import { useCommandKStatus } from '../components/CommandK/useCommandKStatus';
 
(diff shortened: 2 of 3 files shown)


src/entries/popup/hooks/useKeyboardShortcut.ts
```diff
@@ -27,21 +27,12 @@ export function useKeyboardShortcut({
...
   const isCommandKVisible = useCommandKStatus((s) => s.isCommandKVisible);
 
-  // "latest ref pattern" so we don't have to worry about stabilizing the callbacks
-  const handlerRef = useRef(handler);
-  const conditionRef = useRef(condition);
-  useLayoutEffect(() => {
-    handlerRef.current = handler;
-    conditionRef.current = condition;
-  });
-
```
magiziz, line 37: Let's say our `condition` changes at some point, it might be due to a state change for example. The `useRef` will still keep the old reference to the value without it getting updated.

If you were to try to do `condition?.()` you won't get the latest value

magiziz: Same with `handler`

e2e/parallel/shortcuts-home.test.ts
```diff
@@ -216,6 +216,27 @@ describe.runIf(browser !== 'firefox')(
...
+      await executePerformShortcut({ driver, key: 'k' });
+      await executePerformShortcut({ driver, key: 'ENTER' });
+      await checkExtensionURL(driver, 'send');
+    });
+
+    it('should navigate to my wallets section in Cmd+K', async () => {
+      await goToPopup(driver, rootURL);
+      await executePerformShortcut({ driver, key: 'k' });
+      await executePerformShortcut({
+        driver,
```
magiziz, line 229: super simple e2e tests

DanielSinclair: Nice!

DanielSinclair (approved):
@magiziz Backspace fix works, but noticed that the context menu doesn't get dismissed upon `C` for copy (action is triggered though)
Screenshot 2024-05-14 at 4 47 44 PM

DanielSinclair:
Can do another quick review to ensure that dismissal is fixed when you're ready

magiziz:
@DanielSinclair Yep i'll take a look at the "c" shortcut problem 👍

magiziz:
@DanielSinclair Can you pull latest changes from this branch and see if you can still reproduce it ? I'm also trying to break it, but seems like "c" shortcut works as expected for me 🤔

https://github.com/rainbow-me/browser-extension/assets/53529533/1d6e2b4c-4a6d-4988-a365-a82f738d803d

estebanmino (approved):
didnt try it but code lgtm

src/entries/popup/hooks/useKeyboardShortcut.ts
greg-schrammel: can you check if this is not making it subscribe and unsubscribe from the keydown event on every render?
and if so (not sure), do you think of an alternative other than wrapping condition and handler on `useCallback` on use?
I guess the most 'performant' way would be to remember to ALWAYS call this hook like this
```ts
useKeyboardShortcut({
  handler: useCallback(() => {
    if (lalala) a()
    b()
  }, [lalala]),
  condition: useCallback(() => {
    if (lalala) return a()
    return b()
  }, [lalala])
})
```

DanielSinclair:
@magiziz Tokens and NFTs menu look good; still seeing dismissal issue here on Activity

https://github.com/rainbow-me/browser-extension/assets/4412473/4850913f-ee47-4e1d-8a70-d34557a675e5

src/entries/popup/hooks/useKeyboardShortcut.ts
magiziz: Yeah this will subscribe / unsubscribe when `handler` or `condition` changes. That's why we always need to use `useCallback` like you mentioned, otherwise on every re-render the function will be marked as "new / changed" everytime. Here is a good example of how `useKeyboardShortcut` should be used https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/hooks/useTokensShortcuts.ts#L209-L212

I wonder if we should also go through all places in the code where we use `useKeyboardShortcut` and just do `useCallback` or `useMemo` for `condition` and `handler`. Maybe in the next follow up PR that'd be better ?

magiziz:
@DanielSinclair Fixed the dismissal context menu issue on Activity tab

PR #1544, opened 2024-05-16
fix: auto browser language translation for seed phrases / private keys

magiziz:
Fixes BX-1466

Some browser like microsoft edge and google chrome can translate words / letters to another language, this can happen if user wishes to use "translation" feature in the browser setting. The only problem with that is when seed phrases appear they get translated to another language instead of the english language, same with private keys the letters can be translated to another letter language.

**Before (Microsoft edge):**

before

**After (Microsoft edge):**

after

## What to test

- You can use google translate feature in google chrome or use microsoft edge translation feature by going to settings -> languages -> "Offer to translate pages..." and remove english as preferred language and choose any other language which you'd want to translate.
- Try creating a new wallet group and see if seed phrases are not being translated to another language and same with private keys.

### DIFF
--- a/src/design-system/components/Box/Box.tsx
+++ b/src/design-system/components/Box/Box.tsx
@@ -1,6 +1,6 @@
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
 import clsx, { ClassValue } from 'clsx';
-import React, { forwardRef } from 'react';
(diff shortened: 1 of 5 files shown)


DanielSinclair (approved):
god bless 🫡

estebanmino (approved):

PR #1547, opened 2024-05-22
feat: token search (Cmd+K)

magiziz:
Fixes BX-1452, BX-1479
Figma link (if any):

With Cmd+K you can now paste in any token address and search for a token.

https://github.com/rainbow-me/browser-extension/assets/53529533/94f863c3-9145-4584-9f34-a6435107d8ed

## What to test

- Open Cmd+K and paste in token address from any supported chains, ideally it would be best if you can test multiple token addresses from different chains. After you've done that you should be able to see them in Cmd+K menu.
- If you search a token address and you own the same token you should only see one token row in Cmd+K menu, which is your owned one.
- You cannot watch a token address as a wallet. You shouldn't see this:
image

- Make sure whenever you press "Enter" for a searched remote token it should redirect you to the token details page like you normally would with your owned tokens.
- The token details page shouldn't allow you to pin token, hide token and show you current balance and value.
- Make sure token details page for remote tokens works smoothly


DanielSinclair:
@magiziz Tweaked the title here to drop the `custom` piece. Custom Tokens currently refer to the tokens you can add for a network in the Network settings (like MetaMask). I'd call these searched/remote or unowned tokens for clarity

e2e/vitest.config.ts
```diff
@@ -6,7 +6,7 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   test: {
     include: ['./**/**/*.test.ts'],
-    testTimeout: 90_000,
+    testTimeout: 120_000,
```
estebanmino, line 9: did it need more time?

thinking out loud I think this will increase our time waiting for test results by some non trivial amount? 
sometimes we see tests failing 3 times so this will add time

estebanmino (approved):
one minor question

functionality code looks good and working as expected for me

e2e/vitest.config.ts
magiziz: Fixed 👍

src/entries/popup/components/CommandK/utils.ts
```diff
@@ -185,7 +185,7 @@ const memoize = (
       commandListVersion = currentVersion;
     }
 
-    if (!resultsCache.has(query)) {
+    if (!resultsCache.has(query) || commandListVersion === currentVersion) {
```
magiziz, line 188: Got this Cmd+K cache bug, so had to fix it in here

https://github.com/rainbow-me/browser-extension/assets/53529533/3c5d8a88-ee3b-487a-94b2-8facaf8e97ff

magiziz: It didn't sometimes allow me to search for a token

DanielSinclair (approved):

PR #1554, opened 2024-05-28
consolidate chains info

estebanmino:
Fixes BX-####
Figma link (if any):

moving code around to be better prepared when backend sends us chains

- references/chains has now all chains info, but native assets
- utils/chains has all chain related methods, i removed some of them since they were not necessary
- changed places of wagmi util methods to clientRpc now wagmi/index has only wagmi constructor methods
- used viem chains wherever is possible


src/entries/popup/pages/home/NFTs/NFTDetails.tsx
```diff
@@ -804,9 +805,11 @@ const NFTAccordionAboutSection = ({
...
             <Inline alignVertical="center" space="6px">
               <ChainBadge
-                chainId={chainIdFromChainName(
-                  nft?.network || ChainName.mainnet,
-                )}
+                chainId={
+                  chainNameToIdMapping[
+                    (nft?.network as ChainName) || ChainName.mainnet
+                  ]
+                }
```
magiziz, line 812: I assume doing `as` type assertion won't be an issue here right ?

src/entries/popup/pages/send/index.tsx
```diff
@@ -283,14 +283,14 @@ export function Send() {
...
+      chainId: chainNameToIdMapping[nft.network],
       chainName: nft.network,
       isNativeAsset: false,
       name: nft.name,
       symbol: nft.collection.name,
-      uniqueId: `${nft.asset_contract.address || ''}_${chainIdFromChainName(
-        nft.network,
-      )}`,
+      uniqueId: `${nft.asset_contract.address || ''}_${
+        chainNameToIdMapping[nft.network as ChainName]
```
magiziz, line 292: Same here ?

magiziz (approved):
Looks good to me. Just had a few questions.

src/entries/popup/pages/home/NFTs/NFTDetails.tsx
estebanmino: yeah but now that you mention i'll take another look to return that type from the nft interface to make it cleaner

estebanmino: ok, this wasn't needed at all lol i updated

derHowie (approved):
lgtm 🚀

PR #1557, opened 2024-05-30
Reduce internal builds with a commit check

BrodyHughes:
Fixes BX-####
Figma link (if any):

- If the last commit was a version bump, do not do another version bump / internal build.


DanielSinclair (approved):
Pushed up a little tweak to simplify since we only use this in the internal release script. Tested with the runner and lgtm!

magiziz (approved):

PR #1562, opened 2024-06-05
fix: `NaN` shown in tokens tab bar

magiziz:
Fixes BX-1490

- Fixed an issue where `NaN` would show up in tab bar from home page

**Before:**

before

**After:**

after

### DIFF
--- a/src/entries/popup/hooks/useUserAssetsBalance.ts
+++ b/src/entries/popup/hooks/useUserAssetsBalance.ts
@@ -46,7 +46,7 @@ export function useUserAssetsBalance() {
   );
 
   const {
-    data: totalAssetsBalanceCustomNetworks = [],
+    data: totalAssetsBalanceCustomNetworks,
     isLoading: customNetworksIsLoading,
   } = useCustomNetworkAssets(
     {
@@ -65,8 +65,8 @@ export function useUserAssetsBalance() {
   );
 
   const totalAssetsBalance = add(
-    totalAssetsBalanceKnownNetworks as string,
-    totalAssetsBalanceCustomNetworks as string,
+    totalAssetsBalanceKnownNetworks ?? '0',
+    totalAssetsBalanceCustomNetworks ?? '0',
   );
 
   return {
--- a/src/entries/popup/hooks/useUserAssetsBalance.ts
+++ b/src/entries/popup/hooks/useUserAssetsBalance.ts
@@ -64,14 +64,16 @@ export function useUserAssetsBalance() {
     },
   );
 
-  const totalAssetsBalance = add(
-    totalAssetsBalanceKnownNetworks ?? '0',
-    totalAssetsBalanceCustomNetworks ?? '0',
-  );
+  const totalAssetsBalance =
+    totalAssetsBalanceKnownNetworks && totalAssetsBalanceCustomNetworks
+      ? add(totalAssetsBalanceKnownNetworks, totalAssetsBalanceCustomNetworks)
+      : undefined;
 
   return {
     amount: totalAssetsBalance,
-    display: convertAmountToNativeDisplay(totalAssetsBalance || 0, currency),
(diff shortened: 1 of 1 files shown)


estebanmino (approved):
amazing

derHowie (approved):

DanielSinclair:
@magiziz Where does the $Nan value originate? In the screenshot case, $0 would also be the wrong value. I lean to hiding the label when there is no value available; Coinbase users often complain about $0 display values during service outages.

magiziz:
@DanielSinclair Yeah it happens when there the value `undefined`. We can hide it if the value is `undefined`.

DanielSinclair:
@magiziz Gotcha. Let's do that! Can you also report this instance to Backend team? Seems like an issue for that wallet in addys

DanielSinclair (changes requested):
Mentioned above

magiziz:
@DanielSinclair Should be ready now. This is how it'll look if value is `undefined`.

Screenshot 2024-06-07 at 20 18 38

DanielSinclair:
@magiziz Can't seem to ever encounter this scenario. Is there an easy way to reproduce?

src/entries/popup/hooks/useUserAssetsBalance.ts
```diff
@@ -64,14 +64,16 @@ export function useUserAssetsBalance() {
...
+  const totalAssetsBalance =
+    totalAssetsBalanceKnownNetworks && totalAssetsBalanceCustomNetworks
+      ? add(totalAssetsBalanceKnownNetworks, totalAssetsBalanceCustomNetworks)
+      : undefined;
 
   return {
     amount: totalAssetsBalance,
-    display: convertAmountToNativeDisplay(totalAssetsBalance || 0, currency),
+    display: totalAssetsBalance
+      ? convertAmountToNativeDisplay(totalAssetsBalance || 0, currency)
```
DanielSinclair, line 75: I think we can remove the `||` here

src/entries/popup/hooks/useUserAssetsBalance.ts
```diff
@@ -64,14 +64,16 @@ export function useUserAssetsBalance() {
     },
   );
 
-  const totalAssetsBalance = add(
-    totalAssetsBalanceKnownNetworks as string,
-    totalAssetsBalanceCustomNetworks as string,
-  );
+  const totalAssetsBalance =
+    totalAssetsBalanceKnownNetworks && totalAssetsBalanceCustomNetworks
+      ? add(totalAssetsBalanceKnownNetworks, totalAssetsBalanceCustomNetworks)
```
DanielSinclair, line 69: Was the `as string` here legacy, or was there a particular reason we had to convert that type?

magiziz: It was type casted because of the undefined we had. Now we don't need to type cast it.

src/entries/popup/hooks/useUserAssetsBalance.ts
magiziz: Done 👍

src/entries/popup/hooks/useUserAssetsBalance.ts
DanielSinclair: Gotcha

DanielSinclair (approved):

magiziz:
> @magiziz Can't seem to ever encounter this scenario. Is there an easy way to reproduce?

@DanielSinclair It's hard to reproduce this one. You'd need to find an address that holds a lot of assets and then just open and close the BX, and then re-open again. You will see `NaN` for few seconds. This happens to `vitalik.eth` sometimes.

BrodyHughes (approved):
tested and looks good. man huge wallets do take a while to load tho (not your fault).

PR #1564, opened 2024-06-06
feat: fetch networks from be

estebanmino:
Fixes BX-####
Figma link (if any):

with this pr we'll start fetching chains information from backend at build time

- we fetch networks into a json
- we use that file  to build the wagmi client with custom information chains
- we use that file to get additional information as which chain support swap, transactions, etc. `servicesEnabled` from the `BackendNetwork` interface
- we use that info to get native asset address
- i removed a bunch of hardcoded vars related to chains that are now dynamic


scripts/networks.js
```diff
@@ -0,0 +1,107 @@
...
+  try {
+    console.log('Fetching networks ...');
+    await fetchData();
+    console.log('Networks data fetched and available.');
+    process.exit(0);
+  } catch (error) {
+    console.error('Error fetching networks data:', error);
+    process.exit(1);
+  }
+}
```
magiziz, line 105: 🤩

src/core/references/chains.ts
```diff
@@ -1,198 +1,104 @@
...
+
+export const supportedApprovalsChainIds = filterChainIdsByService(
+  (services) => services.addys.approvals,
+);
+
+export const supportedTransactionsChainIds = filterChainIdsByService(
+  (services) => services.addys.transactions,
+);
+
+export const supportedAssetsChainIds = filterChainIdsByService(
```
magiziz, line 90: This is so clean!

src/core/resources/chains/chainMetadata.ts
```diff
@@ -1,14 +1,28 @@
...
 
+export const getChainMetadataRPCUrl = async ({
+  rpcUrl,
+}: {
+  rpcUrl?: string;
+}) => {
+  if (rpcUrl && isValidUrl(rpcUrl)) {
+    const provider = new JsonRpcProvider(proxyRpcEndpoint(rpcUrl, 0));
+    const network = await provider.getNetwork();
+    return { chainId: network.chainId };
```
magiziz, line 21: Is it possible to use viem ?

src/entries/popup/hooks/useWalletName.ts
```diff
@@ -2,10 +2,15 @@ import { Address } from 'viem';
...
+import { ChainId } from '~/core/types/chains';
 import { truncateAddress } from '~/core/utils/address';
 
 export const useWalletName = ({ address }: { address?: Address }) => {
-  const { data: ensName } = useEnsName({ address });
+  const { data: ensName, error } = useEnsName({
+    address,
+    chainId: ChainId.mainnet,
+  });
+  console.log('-- useWalletName', ensName, error);
```
magiziz, line 13: Remove ?

magiziz (approved):
This is awesome! Nothing major just had some few questions 🙌👍

src/core/resources/chains/chainMetadata.ts
estebanmino: in which sense?

this method is used when the user puts an url in the custom network form so we don't have that rpc as a network in wagmi, so this is how we're just fetching the chainid

src/entries/popup/hooks/useWalletName.ts
estebanmino: uh

PR #1570, opened 2024-06-10
fix: bridge quote not working for some tokens

magiziz:
Fixes BX-1495
Figma link (if any):

We don't check if the token is native or non native from destination chain. For example when bridging Degen on degen chain to Degen on base chain the quote will fail. It'll search for `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` instead of the Degen token address on base.

Here is a bug where user can't bridge Degen on degen chain to Degen on base chain

https://github.com/rainbow-me/browser-extension/assets/53529533/96f98e18-7e57-4cc7-b89d-01ccced77831

## What to test

- Make sure you can bridge Degen on degen chain to Degen on base chain

### DIFF
--- a/src/entries/popup/hooks/useSearchCurrencyLists.ts
+++ b/src/entries/popup/hooks/useSearchCurrencyLists.ts
@@ -15,7 +15,7 @@ import {
   TokenSearchThreshold,
 } from '~/core/types/search';
 import { isSameAsset } from '~/core/utils/assets';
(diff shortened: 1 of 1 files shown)


estebanmino (approved):
works! 🚀

magiziz:
@BrodyHughes @DanielSinclair Let me know if it works.

derHowie (approved):

PR #1578, opened 2024-06-17
fix: provider gas

estebanmino:
Fixes BX-1520
Figma link (if any):

for chains that we don't officially support we currently have a provider `getFeeData` call, but it doesn't work very well, specifically foro zksync there's a different mechanism https://docs.zksync.io/zk-stack/concepts/fee-mechanism that. can be resumed using gasprice

so i'm deleting the old `getFeeData` call and only using the old  `getGasPrice`

tx sent https://explorer.zksync.io/tx/0x0fccc2a888c1b6e9be5d70c88c2c87f1717f702b8202c12f2ee1566da208737c

and the fee looks ok now

image

### DIFF
--- a/src/core/resources/gas/gasData.ts
+++ b/src/core/resources/gas/gasData.ts
@@ -43,6 +43,9 @@ export const useGasData = ({ chainId }: { chainId: ChainId }) => {
       },
     );
 
+  console.log('meteorologyData', meteorologyData);
+  console.log('providerGasData', providerGasData);
+
   return {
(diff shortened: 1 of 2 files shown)


magiziz (approved):
Nice!
Just a curious question, why did we use `provider.getFeeData()` instead of `provider.getGasPrice()` ?

DanielSinclair (approved):

DanielSinclair (approved):

PR #1580, opened 2024-06-18
fix: home shortcuts not unregistered in token details

magiziz:
Fixes BX-1486
Figma link (if any):

- Home shortcuts weren't unregistered in token details page. Technically it should be unregistered since token details  page is a whole new route / component, but we used token details page as a child route of the home route which didn't unmount the home shortcut listener.

## What to test

- Make sure when you land in token details page you can copy address by pressing `C` shortcut and same with `P` for pin.


estebanmino (approved):

BrodyHughes (approved):
🫡

PR #1579, opened 2024-06-17
feat: token search on chain (Cmd+K, swap and token details)

magiziz:
Fixes BX-1504, BX-1369
Figma link (if any):

## What changed (plus any additional context for devs)

Backend token search API doesn't return all tokens. If we don't detect any tokens from the backend API, we'll just search them on chain. This is now supported in swap page, Cmd+K menu and token details page.

## Screen recordings / screenshots

https://github.com/rainbow-me/browser-extension/assets/53529533/05078f32-dad5-4a2d-a3c4-6751117c4a04

## What to test

- Search for a random non supported backend token in swap page, and make sure the swap flow works.
- Open Cmd+K menu and search for a non supported backend token. You should be able to see it, and also have the ability to go to the token details page.

_You can easily see which token is not supported from backend by searching any contract address in swap page (production) and do the same thing in this branch and compare._

_For example a token that's not supported by the backend API is $PALAI (0x3Cb48AeB3D1abaDc23d2D8a6894B3A68338381C2)._

PR #1584, opened 2024-06-19
feat: remote token search

magiziz:
Fixes BX-1505
Figma link (if any):

**IMPORTANT!** Merge [this](https://github.com/rainbow-me/browser-extension/pull/1579) PR first, before merging this. Since there are some bugs that were fixed + there will be merge conflicts.

We now support remote token search for verified assets and unverified assets in Cmd+K and unverified assets in token details page.

https://github.com/rainbow-me/browser-extension/assets/53529533/efd3113a-1172-466e-8d0c-4707cf38fcf6

### Cmd+K:

- Type at least  **three** characters to search for a token.
- None of the  **unowned** tokens should have a badge on the right.
- You should see a **loading bar** in the Cmd+K menu if tokens are being searched.
- Try searching for your **owned** asset (e.g., "Ethereum" if you have ETH) and make sure that:
  - **Owned** assets are ranked as priority 1.
  - **Verified** assets are ranked as priority 2.
  - **Unverified** assets are ranked as priority 3.
  - You can see which assets are **verified** and **unverified** by looking at:
    - name (make sure the name doesn't sound like a scam if it's a verified asset)
    - icon (verified assets will have an icon)

### Token details page:

- If you select an unverified asset in Cmd+K you should be redirected to token details page and see the token details.


estebanmino:
@magiziz what do you mean by `should have a badge on the right`? none of my assets have that, even owned assets

following question, for mainnet assets there's no difference between owned and unowned assets, is that ok?

magiziz:
> @magiziz what do you mean by should have a badge on the right? none of my assets have that, even owned assets

@estebanmino Ah yeah those are confusing. The badges i'm talking about are these:

image

If you have owned token they badges should appear, but search tokens (unowned) shoudn't have any badge.


> following question, for mainnet assets there's no difference between owned and unowned assets, is that ok?

I think mainnet assets can be owned and unowned depending on what you have. For example if i own ethereum on base that's an owned asset, but if i search for ethereum on optimism and don't have that in my wallet we can consider it as unowned asset, but happy to do any changes. Lmk what you think.

estebanmino (approved):
lgtm!

BrodyHughes (approved):
tested all requested flows, lgtm.

PR #1599, opened 2024-06-25
fix: BX-1540

estebanmino:
Fixes BX-1540
Figma link (if any):

fixing ticket, seems like we had a something not configured well in inpage when coinbase wallet is present, not really sure why but this fixes the issue

maybe was this line https://github.com/rainbow-me/browser-extension/pull/1599/files#diff-d13d2b4eaccf47cdf4ab7013649e5884d9a92832fe986e41e5dcad32f9c1e031L149 trying to set a freezed value, but is weird that is not an issue for most of users

### DIFF
--- a/src/entries/inpage/index.ts
+++ b/src/entries/inpage/index.ts
@@ -21,7 +21,7 @@ declare global {
     lodash: unknown;
     rainbow: RainbowProvider;
     providers: RainbowProvider[];
-    walletRouter: {
+    rnbwWalletRouter: {
       rainbowProvider: RainbowProvider;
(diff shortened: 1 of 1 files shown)


magiziz (approved):
🚀

DanielSinclair (approved):
Adding approval for now @estebanmino. Sent @magiziz some notes about if there would be any side effects of `walletRouter` name changing in case any other wallet injectors are expecting to use this, so he will test additional wallets to evaluate. But it is possible we could deprecate this entirely to simplify the implementation now that 6963 is widespread.

PR #1600, opened 2024-06-26
Fix stimmy rounding pt 2

BrodyHughes:
Fixes BX-1541
Figma link (if any):

- couple other places in the flow where the decimals were not rounded at 6 spots

BEFORE:
Screenshot 2024-06-26 at 11 49 35 AM
AFTER:
Screenshot 2024-06-26 at 11 49 35 AM

### DIFF
--- a/src/entries/popup/pages/home/Points/PointsDashboard.tsx
+++ b/src/entries/popup/pages/home/Points/PointsDashboard.tsx
@@ -892,6 +892,11 @@ function MyEarnings({ earnings = '0' }: { earnings?: string }) {
     decimals: 18,
     symbol: eth?.nativeAsset?.symbol,
   });
+  const earningsWithThreshold = handleSignificantDecimalsWithThreshold(
+    convertRawAmountToDecimalFormat(earnings, 18),
+    6,
+    '0.000001',
+  );
   return (
     <Card borderColor="separatorSecondary" display="flex">
       <Stack space="20px">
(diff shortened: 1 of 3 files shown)


src/entries/popup/pages/home/Points/PointsDashboard.tsx
```diff
@@ -892,6 +892,11 @@ function MyEarnings({ earnings = '0' }: { earnings?: string }) {
     decimals: 18,
     symbol: eth?.nativeAsset?.symbol,
   });
+  const earningsWithThreshold = handleSignificantDecimalsWithThreshold(
+    convertRawAmountToDecimalFormat(earnings, 18),
+    6,
+    '0.000001',
```
estebanmino, line 898: why with threshold and not just `handleSignificantDecimals`?

BrodyHughes: I saw them both and just found this easier to understand at first so i used it haha. Is there a significant difference between the two where i shouldn't use withThreshold?

DanielSinclair:
cc @derHowie for another review

DanielSinclair (approved):

src/entries/popup/pages/home/Points/PointsDashboard.tsx
estebanmino: is just one additional param that i don't think we need but won't block

estebanmino (approved):

PR #1606, opened 2024-07-02
chore: add additional custom networks

DanielSinclair:
Fixes BX-####
Figma link (if any):

- added new network icons https://github.com/rainbow-me/assets/pull/91
- added new prepopulated custom networks
- filtering out testnets from network autofill when dev mode is disabled


src/entries/popup/pages/settings/customChain/index.tsx
```diff
@@ -92,18 +93,29 @@ const KNOWN_NETWORKS: { name: string; networkInfo: customNetworkInfo }[] = [
     },
   },
   {
-    name: 'Blast Testnet',
+    name: 'Aurora Testnet',
```
estebanmino, line 96: would it be a good moment to try to do a loop through viem/chains if is possible, so we don't add more hardcoded code like this?

estebanmino (approved):

src/entries/popup/pages/settings/customChain/index.tsx
DanielSinclair: @estebanmino I think we'll be able to pull these from the Backend soon. I'm sharing these changes with Benjamin so that he can see what we hardcode currently

DanielSinclair:
@BrodyHughes When you get a second to review; simple networks additions and now filtering testnets when dev mode is off

derHowie (approved):

PR #1613, opened 2024-07-03
fix: window. rainbow providers

estebanmino:
Fixes BX-####
Figma link (if any):

`window.rainbow.providers` was only defined once at the beginning, so any updates to `rnbwWalletRouter.providers` won't be reflected in `window.rainbow.providers`, in which case were rainbow provider is `window.ethereum` the `window.ethereum.providers` would only have the first definition of `window.rainbow.providers` at the beginning 

EDIT

ended up removing `isMetaMask` flag when `announceProvider`

### DIFF
--- a/lavamoat/build-webpack/policy.json
+++ b/lavamoat/build-webpack/policy.json
@@ -1156,8 +1156,13 @@
         "define": true
       },
       "packages": {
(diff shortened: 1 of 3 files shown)


magiziz:
When testing with [manifold](https://www.manifold.xyz/) and [1inch](https://app.1inch.io/#/1/simple/swap/ETH) it seems like giving me these errors:

image

image

BrodyHughes (approved):
tested this on sushiswap and warpcast

magiziz (approved):
lgtm

PR #1614, opened 2024-07-03
fix: delete `isMetaMask` property and share same EIP-1193 instance with EIP-6963

magiziz:
Fixes BX-####
Figma link (if any):

- `window.ethereum.providers` causing infinite loop with Rainbow wallet when having Coinbase and MetaMask installed. This is causing a memory leak due to `window.ethereum.providers` overrides.

EDIT:
This is a small improvement from previous fix we did

### DIFF
--- a/src/entries/inpage/index.ts
+++ b/src/entries/inpage/index.ts
@@ -69,12 +69,6 @@ const rainbowProvider = new RainbowProvider({
 delete rainbowProvider.isMetaMask;
 
 if (shouldInjectProvider()) {
-  // eslint-disable-next-line prefer-object-spread
-  const providerCopy = Object.create(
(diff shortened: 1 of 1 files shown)


magiziz:
This PR is an alternative to https://github.com/rainbow-me/browser-extension/pull/1613#pullrequestreview-2157621157

DanielSinclair:
@magiziz Esteban's PR includes a `isMetaMask = false` now for the EIP-6963 announcer. Does that accomplish the same thing as this PR?

magiziz:
Seems like e2e test fails because we use this e2e app https://bx-e2e-dapp.vercel.app. It requires `isMetaMask` to be true.

PR #1618, opened 2024-07-08
fix: BX-1513

estebanmino:
Fixes BX-1513
Figma link (if any):

adding simulateTransactionsWithoutGas to be used with useSimulateTransaction in dapp tx request


DanielSinclair (approved):
Not sure I was able to replicate the exact scenario, but transaction simulation, gas estimates, and signing lgtm

derHowie (approved):

PR #1626, opened 2024-07-16
fix: BX-1548

estebanmino:
Fixes BX-1548
Figma link (if any):

adding 10% buffer to improve this case


DanielSinclair (approved):
Saw 1 out of gas error after a bunch of testing, but does seem to improve it a lot

magiziz (approved):

PR #1628, opened 2024-07-18
chore: reduce artifact noise

DanielSinclair:
Fixes BX-####
Figma link (if any):

- deprecate artifact comment links to reduce noise
- leverage persistent artifacts on workflow runs instead


derHowie (approved):

estebanmino:
@DanielSinclair build step is failing here

DanielSinclair:
I realized Workflow runs will already persit builds and e2e failure screenshots, so I think we can safely deprecate these additional PR comments all together
Screenshot 2024-08-01 at 5 50 35 PM

magiziz (approved):

PR #1631, opened 2024-07-19
feat: custom gas in all networks

estebanmino:
Fixes BX-1546, BX-1507
Figma link (if any):

enabling custom gas in all networks for all txs types

https://www.loom.com/share/19d39fed3c1040c29ae00a41eddc8a81

### DIFF
--- a/src/entries/popup/components/TransactionFee/CustomGasSheet.tsx
+++ b/src/entries/popup/components/TransactionFee/CustomGasSheet.tsx
@@ -194,7 +194,6 @@ export const CustomGasSheet = ({
   const maxPriorityFeeInputRef = useRef<HTMLInputElement>(null);
   const gasPriceInputRef = useRef<HTMLInputElement>(null);
 
-  console.log('--- customSpeed', customSpeed);
   const [maxBaseFee, setMaxBaseFee] = useState(
     (customSpeed as GasFeeParams)?.maxBaseFee?.gwei,
   );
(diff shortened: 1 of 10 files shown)


src/entries/popup/components/TransactionFee/CustomGasSheet.tsx
```diff
@@ -196,12 +201,19 @@ export const CustomGasSheet = ({
...
+
   const [maxBaseFeeWarning, setMaxBaseFeeWarning] = useState<
     'stuck' | 'fail' | undefined
   >(undefined);
   const [maxPriorityFeeWarning, setPriorityBaseFeeWarning] = useState<
     'stuck' | 'fail' | undefined
   >(undefined);
+  const [gasPriceWarning, setGasPriceWarning] = useState<
+    'stuck' | 'fail' | undefined
+  >(undefined);
```
magiziz, line 216: Maybe we can do something like this ?

```tsx
const [gasPriceWarning, setGasPriceWarning] = useState<'stuck' | 'fail'>();
```

src/entries/popup/hooks/useGas.ts
```diff
@@ -204,6 +218,43 @@ const useGas = ({
...
+  useEffect(() => {
+    if (
+      !gasData ||
+      !enabled ||
+      prevDebouncedGasPrice === debouncedGasPrice ||
+      feeType !== 'legacy' ||
+      !nativeAsset
+    ) {
+      return;
+    }
```
magiziz, line 230: Can we use `useQuery` here ?

estebanmino:
@magiziz I just applied your comments! 🙏

derHowie (approved):

BrodyHughes:
Functionality seems good to me. FOr this sort of case we should do a '<0.0001' or something though:

Screenshot 2024-08-01 at 4 06 23 PM

DanielSinclair:
@BrodyHughes Which network was that? Wonder if it is a bad result from Backend

DanielSinclair (approved):

DanielSinclair:
Lgtm. Have a few design tweaks for existing Custom Gas UI problems, so will create a seperate polish ticket for that and carry over Brody's issue

PR #1634, opened 2024-07-22
import screen doesn't show account balance in USD shows $0

derHowie:
Fixes BX-1447

Fixed a bug that caused wallet import info to display a mainnet balance of $0.

Screenshot 2024-07-22 at 6 43 03 AM

## What to test
Ensure that the mainnet balance is not excluded from the wallet import row.


magiziz (approved):

src/entries/popup/hooks/useWalletsSummary.ts
```diff
@@ -61,7 +61,7 @@ const parseAddressSummary = ({
   }).amount;
   const ethCurrencyBalance = convertAmountAndPriceToNativeDisplay(
     ethBalance || 0,
-    nativeAssets?.[`${ETH_ADDRESS}_1`]?.price?.value || 0,
+    nativeAssets?.[ChainId.mainnet]?.price?.value || 0,
```
estebanmino, line 64: we have other chains below following the same pattern, shouldn't we change it for all networks?

DanielSinclair:
@derHowie Noticed Esteban's comment here; do you need help migrating these other chain ids, or was mainnet just treated differently?

derHowie:
> @derHowie Noticed Esteban's comment here; do you need help migrating these other chain ids, or was mainnet just treated differently?

Just checked. His suggestion was correct. I have included the change.

PR #1640, opened 2024-07-24
fix: disable activity details shortcuts without dropdown

derHowie:
Fixes BX-1565

From ticket:
"Goal:

Only register Token Details shortcuts when the context menu is visibile

Allow a user to select text and copy"

The shortcuts associated with the more options menu in token details now are only active when the dropdown is visible.

## What to test
Make sure that token details does not intercept the 'C' key when the more options dropdown is hidden.


DanielSinclair (approved):
Perfect 👌

magiziz (approved):

PR #1645, opened 2024-07-29
ledger clear signing upgrades

DanielSinclair:
Fixes BX-####
Figma link (if any):

- upgraded `@ledgerhq/hw-app-eth` and `@ledgerhq/hw-transport-webhid`
- removed patch and `static/json/ledger` dynamically imported files

## Follow ups
- evaluate alternative performance optimizations after removing dynamic json imports
- create a wrapper around `@ledgerhq/` packages with a dynamic runtime module linking so that we can import the Ledger module and JSON files as needed, without creating overhead on boot or for wallets that don't use a Ledger device. This will be easier to coordinate more frequent Ledger SDK upgrades required for Clear Signing

## What to test
- Performance of boot/signing before and after ledger upgrades (with or without device paired)
- Pairing and Signing with Ledger devices. Should try on a Ledger with an older firmware version and a recent firmware version (this month)
- Signing for popular dApps like Lido or Uniswap should not display a Blind Signing warning (new aggressive warning)
- Bruno: Firefox bundle size might be too large; need to check sizes or disable bundling for Firefox. Should discuss bundle size improvements with Ledger.


BrodyHughes:
Tried a swap and got stuck here:
Screenshot 2024-07-31 at 1 42 51 PM
Tried a send and got stuck after hitting the send action button. Will dm u the console logs.

DanielSinclair:
Moving to draft until we make progress with Ledger SDK issues

PR #1647, opened 2024-07-30
fix: BX-1582

estebanmino:
Fixes BX-1582
Figma link (if any):

- fixing original ticket, description of what i did on linear
- I also fixed an issue where you can't remove custom rpcs from rainbow supported chains, now you can remove all rpcs but default's rainbow rpc

https://www.loom.com/share/f89332928aa14924b444fc2b957cadc2


src/entries/background/handlers/handleProviderRequest.ts
```diff
@@ -295,11 +295,10 @@ export const handleProviderRequest = ({
         };
         const rainbowChain = rainbowChains[chainObject.id];
         const alreadyAddedRpcUrl = rainbowChain.chains.find(
-          (chain: Chain) =>
-            chain.rpcUrls.default.http[0] === rpcUrl &&
-            rainbowChain.activeRpcUrl === rpcUrl,
+          (chain: Chain) => chain.rpcUrls.default.http[0] === rpcUrl,
```
estebanmino, line 298: checking if the rpcurl was added before, so we don't set it as active

src/entries/popup/pages/settings/rpcs.tsx
```diff
@@ -254,7 +254,8 @@ export function SettingsNetworksRPCs() {
                   <ContextMenu>
                     <ContextMenuTrigger
                       disabled={
-                        mainnetChains[index].name === supportedChain?.name
+                        supportedChain?.rpcUrls.default.http[0] ===
+                        chain.rpcUrls.default.http[0]
```
estebanmino, line 258: the logic we had here was disabling the option to remove the rpc url of rainbow supported chains, now we'll only block the removal of the rainbow rpc

magiziz (approved):

DanielSinclair (changes requested):
Spotted a few issues:
- Noticed that the network toggle will flip to disable when removing a 2nd RPC (I suppose the duplicate RPC URLs match so the network will get caught in a check based on that logic).
- It seems like we can remove the testnet RPCs now as well; any way to limit that?

Tested against Zora and didn't see a new duplicate get added after I cleared that network 👍 

I do see duplicates across other networks as well. Should we do another migration to clear out those duplicates once this logic change is settled?
Screenshot 2024-07-31 at 2 32 45 AM

estebanmino:
@DanielSinclair how did you get those duplicated networks? i don't have any

estebanmino:
@DanielSinclair i also fixed those issues you mentioned with these changes, they're are currently in prod

DanielSinclair (approved):
@estebanmino Changes lgtm. I'm not quite sure where the duplicates came from. I have them on my internal build but not prod, so it must have been one of the migrations; not sure if it impacts prod in other scenarios.

PR #1612, opened 2024-07-03
Rounding large numbers

BrodyHughes:
Fixes BX-1550, BX-121, BX-1551

## What changed:
- this PR effects the swap review cards and the swap review
- follow up PRs to follow changing this behavior elsewhere in the app
- numbers >= 1,000 are now abbreviated using K, M, B, T, etc.

## Screen recordings / screenshots
New behavior:
Screenshot 2024-07-02 at 11 07 26 PM
Screenshot 2024-07-02 at 11 05 58 PM
Screenshot 2024-07-02 at 11 08 23 PM
Screenshot 2024-07-09 at 3 46 58 PM

## What to test
- do large and small values round correctly on the review cards

PR #1648, opened 2024-07-30
fix: text overflow on approvals section

magiziz:
Fixes BX-1522
Figma link (if any):

If the approval amount is too long it might cause overflow issues.

Before:

overflow

After:

approvals_after

## What to test

Make sure text doesn't overflow for large amount approvals


DanielSinclair (changes requested):
I think we would actually want to adopt rounding here so that you can still see the token symbol; it would also support very large numbers. We should be able to use the same logic from this PR: https://github.com/rainbow-me/browser-extension/pull/1612

magiziz:
@DanielSinclair Done 👍

estebanmino (approved):

DanielSinclair (approved):

PR #1649, opened 2024-08-01
perceived finality

derHowie:
Fixes BX-1580
tech spec: https://www.notion.so/rainbowdotme/Perceived-Finality-doc-12d4f7e088e74654ab433d5c661de220

When a pending transaction is detected as confirmed, we are flagging the affected assets' addresses and refetching updated user assets balances from BE.

Additionally, pending transactions were being almost immediately removed from the UI due to an existing bug. The pending transactions should now remain for the full duration.

## What to test
Ensure the `token` params sent to BE reflect the correct addresses when the app attempts to update user asset balances.

Suggestions from Jin:
- multiple transactions, different assets
- multiple transactions, impacting same assets
- Swap 1: ETH → USDC
- Swap 2: SHITCOIN → ETH


estebanmino (commented):
in BX we already have a state that does something similar to staleBalances

src/core/state/swapAssetsToRefresh/index.ts

can we merge both together? just so we don't have dead code, swapAssetsToRefresh gets balances from provider

src/entries/popup/hooks/useWatchPendingTransactions.ts
```diff
@@ -43,25 +43,20 @@ export const useWatchPendingTransactions = ({
...
 
   const pendingTransactions = useMemo(
     () => storePendingTransactions[address] || [],
     [address, storePendingTransactions],
   );
 
-  const refreshAssets = useCallback(
-    (tx: RainbowTransaction) => {
-      if (tx.type === 'swap') {
-        swapRefreshAssets(tx.nonce);
```
estebanmino, line 55: ok i think we need to remove that state slice?

src/entries/popup/hooks/useWatchPendingTransactions.ts
```diff
@@ -244,17 +239,36 @@ export const useWatchPendingTransactions = ({
...
-    if (minedTransactions.length) {
-      await queryClient.refetchQueries({
-        queryKey: consolidatedTransactionsQueryKey({
-          address,
-          currency: currentCurrency,
-          userChainIds: Object.keys(userChains).map(Number),
-        }),
-      });
-    }
+    console.log('MINED TX: ', minedTransactions);
```
estebanmino, line 242: log

src/entries/popup/hooks/useTransactionListForPendingTxs.ts
```diff
@@ -106,15 +112,60 @@ function watchForPendingTransactionsReportedByRainbowBackend({
...
+  pendingTransactions.forEach((tx) => {
     const txNonce = tx.nonce || 0;
     const latestTx = latestTransactions.get(tx.chainId);
     const latestTxNonce = latestTx?.nonce || 0;
     // still pending or backend is not returning confirmation yet
     // if !latestTx means that is the first tx of the wallet
-    return !latestTx || txNonce > latestTxNonce;
+    const newlyConfirmed = latestTxNonce && txNonce <= latestTxNonce;
+    if (newlyConfirmed) {
+      newlyConfirmedTransactions = [...newlyConfirmedTransactions, tx];
```
estebanmino, line 126: took me some time to follow this, why not `.push` so `newlyConfirmedTransactions` can be const ?

derHowie: alright, done

src/core/raps/execute.ts
```diff
@@ -177,11 +182,14 @@ export const walletExecuteRap = async (
         };
         const { hash } = await executeAction(actionParams);
         hash && (await waitForNodeAck(hash, wallet.provider));
+        if (index === actions.length - 1) {
```
brunobar79, line 185: What's the logic for this check?

src/core/state/staleBalances/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+    clearExpiredData: (address: Address) => {
+      set((state) => {
+        const { staleBalances } = state;
+        const staleBalancesForUser = staleBalances[address] || {};
+        const newStaleBalancesForUser: StaleBalancesByChainId = {
+          ...staleBalancesForUser,
+        };
+        for (const c of Object.keys(staleBalancesForUser)) {
+          const chainId = parseInt(c);
```
brunobar79, line 80: Gotta add the base to avoid surprises

src/core/state/staleBalances/index.ts
```diff
@@ -0,0 +1,127 @@
...
+          },
+        };
+      });
+    },
+    getStaleBalancesQueryParam: (address: Address) => {
+      let queryStringFragment = '';
+      const { staleBalances } = get();
+      const staleBalancesForUser = staleBalances[address];
+      for (const c of Object.keys(staleBalancesForUser)) {
+        const chainId = parseInt(c);
```
brunobar79, line 107: Same

brunobar79 (approved):
Good stuff!!!

src/core/raps/execute.ts
derHowie: it returns tx hash of the final action in the array. Realizing now that I needed this in my original implementation but don't anymore. I can remove returning this if you want, but I wouldn't be surprised if having the tx hash of the last action in a rap ends up being useful.

greg-schrammel (approved):

estebanmino (approved):

PR #1650, opened 2024-08-02
fix: tabs messenger not responding on fast refresh

magiziz:
Fixes BX-1478
Figma link (if any):

`tabMessenger` now sends message events to all tabs if RPC methods are supported. Before it could send to the active tab only.

**Before**

https://github.com/user-attachments/assets/3015e04b-235d-45dd-aec9-c7eda3a4d2e0

**After**

https://github.com/user-attachments/assets/ac034181-8291-4970-a5f4-112f55031a87

## What to test

- Make sure you open 5 tabs with this url https://blur.io/collections and connect your wallet. Then do fast refresh on every tab and meanwhile switch back and fourth. The wallet should be connected on all tabs.

### DIFF
--- a/src/core/messengers/internal/tab.ts
+++ b/src/core/messengers/internal/tab.ts
@@ -1,3 +1,5 @@
+import { RequestArguments } from '@rainbow-me/provider';
+
 import {
   CallbackFunction,
   ReplyMessage,
@@ -9,7 +11,10 @@ import { isValidSend } from './isValidSend';
 
 let activeTab: chrome.tabs.Tab;
 
-function getActiveTabs() {
+const shouldNotifyAllTabs = (method: string) =>
+  ['eth_requestAccounts'].includes(method);
+
+function getCurrentActiveTab() {
   if (!chrome.tabs) return Promise.resolve([]);
   return chrome.tabs
     .query({ active: true, lastFocusedWindow: true })
@@ -18,6 +23,11 @@ function getActiveTabs() {
       activeTab = tab;
       return [tab];
     });
+}
+
+function getAllActiveTabs() {
+  if (!chrome.tabs) return Promise.resolve([]);
+  return chrome.tabs.query({});
 }
 
 function sendMessage<TPayload>(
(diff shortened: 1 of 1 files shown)


derHowie (approved):

estebanmino (approved):

magiziz:
@BrodyHughes Can you do QA before we merge this ? You can try opening many tabs and just test simple BX behaviours and make sure they work fine. Behaviours which i'm talking about are like sign message, connect wallet, switch chain and etc.

DanielSinclair:
@magiziz Haven't fully tested yet, but thinking through the implications here. I think the expected flow would look like:
1. Open Tab A, Connect Wallet A
2. Open Tab B, Wallet A from Tab A is connected by default, Switch Wallet to Wallet B, Tab A remains connected to Wallet A
3. Request tx from Tab A, transactions are routed accordingly based on current tab session
Tab A and B would be different sessions, so you can use different wallets simultaneously, and transactions are routed accordingly. I think a previous issue is that we treat the dApp URL as a routing mechanism, rather than a unique session id for each tab. Lmk if you think this change is aligned with this behavior

magiziz:
@DanielSinclair I see. I'm able to reproduce this behaviour as well, but this is also happening in production right now.


https://github.com/user-attachments/assets/bc7656a8-7300-4805-b9b2-0c7ecc97fc13

The change is not made in this PR, but we could do a follow up PR that routes the transactions accordingly. What do you think ?

DanielSinclair (changes requested):
@magiziz I think these changes could actually worsen what was reported in the ticket because it will now more reliably send provider message to the wrong tabs; I think we'll need to dive deeper into this and rework the routing. The sign transaction button prompting a signature with a different wallet than the connected wallet is a blocker.

magiziz:
@DanielSinclair I think the changes didn't cause the provider to send messages to the wrong tab. The video above was actually using the production BX version which already had that behaviour and we can have another PR which fixes that issue. Were there any issues you saw with this change ?

The current issue was that when the dApp was requesting something from BX it didn't give back the response since our tab messenger were at freeze mode.

For the current change the only thing i did was to notify all tabs if one of these RPC methods were requested: `eth_requestAccounts`, `eth_accounts` and `eth_chainId`.

magiziz:
@DanielSinclair I think the changes didn't cause the provider to send messages to the wrong tab. The video above was actually using the production BX version which already had that behaviour and we can have another PR which fixes that issue. Were there any issues you saw in this PR ?

The current issue was that when the dApp was requesting something from BX it didn't give back the response since our tab messenger were at freeze mode.

For the current change the only thing i did was to notify all tabs if one of these RPC methods were requested: `eth_requestAccounts`, `eth_accounts` and `eth_chainId`.

magiziz:
Closing this PR since we'll do a follow up refactor for our messengers.

PR #1656, opened 2024-08-07
round swap input and output values

magiziz:
Fixes BX-1549
Figma link (if any):

Rounded swap input and output values

https://github.com/user-attachments/assets/c516afa3-3e51-465d-b413-5051f527e684

## What to test

1. Go to swaps, type in the any maximum amount and make sure the the text doesn't get clipped. The goal here is to make sure the maximum number of numeric values are set to 12 including `.`.
2. Full unrounded values should be available on input hover with the tooltip.


DanielSinclair (changes requested):
Noticed that Bridge routes via the Bridge button seem to escape the rounding
Screenshot 2024-08-07 at 3 58 01 PM
Can we also left align the tooltip to the start of the number like this balance (reversed); noticed that it looks a bit funny when it is centered and you have an exact balance like `20.045 USDC`
Screenshot 2024-08-07 at 3 49 27 PM

magiziz:
@DanielSinclair Pushed a change. Lmk if it looks good.

magiziz:
@greg-schrammel Yeah we tried to show the full rounded amount with last decimal with tooltip when hovering on the input field. Otherwise the input value will be set max to 12 characters.

BrodyHughes (approved):

DanielSinclair:
I found a few edge cases; I think we'll need to refactor this in a follow-up to round based on the horizontal space available. 
Screenshot 2024-08-14 at 1 55 44 AM
Screenshot 2024-08-14 at 1 45 44 AM

DanielSinclair (approved):

DanielSinclair:
@magiziz If we're still stuck on the Tooltips for test reliability here, I think it's fine if we drop those for now.

magiziz:
@DanielSinclair I fixed the tooltip issue and got chrome e2e swap working, but seem to have an issue with firefox e2e send. Will fix this and ping once it's resolved.

DanielSinclair:
@magiziz Disabled Firefox e2e for the time being, so these tests should pass. If you still think there is an explanation for the failing tests, might be worth fixing anyway for when we turn them back on next week (more limited coverage)

DanielSinclair:
@magiziz Want to revert the e2e changes and tweaks and proceed with merge?

magiziz:
@DanielSinclair Yeah the e2e changes and tweaks is removed, i did force push. Will get the e2e passed and get the PR merged.

PR #1657, opened 2024-08-07
fix: proxy

estebanmino:
Fixes BX-1589
Figma link (if any):

Details and repro steps in ticket

The issue was that we weren't proxying the custom rpc since it was a default rainbow chain id, now we'll proxy all chain rpcs that aren't rainbow's rpcs

### DIFF
--- a/lavamoat/build-webpack/policy.json
+++ b/lavamoat/build-webpack/policy.json
@@ -1156,8 +1156,13 @@
         "define": true
       },
       "packages": {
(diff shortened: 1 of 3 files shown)


DanielSinclair (approved):
Tested with a custom RPC on base with in-app swap + uniswap dapp transaction

BrodyHughes (changes requested):
Seems to work on mainnet with custom RPC with `https://eth.llamarpc.com` but it didn't work for me on OP with `https://optimism.llamarpc.com`

Hit the same error as seen in the ticket.

BrodyHughes (approved):
previous issues were rate limiting by llama. looks good!

PR #1660, opened 2024-08-12
limit percentage reduction in token details

magiziz:
Fixes BX-1564
Figma link (if any):

Capped the percentage on the token details page to not decrease more than `-100%`.

**Before:**

BX-1564 (after)

**After:**

BX-1564 (before)

## What to test

Select a token that experienced a significant decrease over time such as $DEGEN and make sure that when hovering over the graph at its lowest point the decrease does not exceed -100%.


estebanmino (approved):
i'm gonna approve but this seems like a bigger issue? not sure if from BE or if we calculate this number

magiziz:
@estebanmino Yeah could be. I looked at our percentage logic [here](https://github.com/rainbow-me/browser-extension/blob/master/src/entries/popup/pages/home/TokenDetails/PriceChart.tsx#L248-L249) and the calculation is done correctly. I think $DEGEN token just had a significant drop.

greg-schrammel (changes requested):
pretty sure the `percentDiff` fn is wrong, we should fix that instead, price -100% and no being 0, is just wrong

magiziz:
We've found out that that there is a small bug in the charts where it shows `Infinity`. I've messaged the backend team to let them know.

DanielSinclair:
Moved to draft while we workout the Backend change

DanielSinclair:
Closing for now because I assume Backend will handle this fix

greg-schrammel (approved):

PR #1665, opened 2024-08-13
Change chrome-e2e-parallel to use beefy-runner-bx

BrodyHughes:
Fixes BX-1601

We currently don't use our BEEFIEST machine so this changes that


greg-schrammel (approved):
🍔

brunobar79 (approved):

PR #1666, opened 2024-08-13
remove ff from CI, move some things around

BrodyHughes:
Fixes BX-1611
Trying to keep these CI tickets in digestable chunks.

Goal: Reduce failure rate on CI checks
- Extract onboarding into its own suite 
- Reincorporate onboarding in main e2e chrome suite 
- Remove firefox:e2e from CI checks on PRs


DanielSinclair:
@BrodyHughes Few pieces of feedback:
- We actually want to keep the Firefox runner for the sanity checks, but only run the Firefox suite (just Onboarding tests) on merges to master. We'd can do this by creating a separate Workflow file that is triggered by:
```
on:
  push:
    branches: [master]
 ```
- I wonder if we should try to preserve the onboarding tests in the existing chrome suites with a reusable job, rather than creating another test. The existing suites would fail if that test would fail, so they're fairly dependent on onboarding passing. We could then create seperate sanity check for chrome, firefox, etc. using that same reusable job/workflow for the merge to master workflow
- We might want to preserve the Firefox handholding checks that exist in case we want to expand the sanity testing; can we comment out or write a comment where we previously would skip a test on Firefox?

BrodyHughes:
> @BrodyHughes Few pieces of feedback:
> 
> * We actually want to keep the Firefox runner for the sanity checks, but only run the Firefox suite (just Onboarding tests) on merges to master. We'd can do this by creating a separate Workflow file that is triggered by:
> 
> ```
> on:
>   push:
>     branches: [master]
> ```
> 
> * I wonder if we should try to preserve the onboarding tests in the existing chrome suites with a reusable job, rather than creating another test. The existing suites would fail if that test would fail, so they're fairly dependent on onboarding passing. We could then create seperate sanity check for chrome, firefox, etc. using that same reusable job/workflow for the merge to master workflow
> * We might want to preserve the Firefox handholding checks that exist in case we want to expand the sanity testing; can we comment out or write a comment where we previously would skip a test on Firefox?

I thought we said we wanted to remove FF completely from e2e, then only run those on merge commits to master? Ticket says:
- Remove firefox:e2e from CI checks on PRs
...
- Add onboarding firefox merge commits on master

BrodyHughes:
> @BrodyHughes Few pieces of feedback:
> 
> * We actually want to keep the Firefox runner for the sanity checks, but only run the Firefox suite (just Onboarding tests) on merges to master. We'd can do this by creating a separate Workflow file that is triggered by:
> 
> ```
> on:
>   push:
>     branches: [master]
> ```
> 
> * I wonder if we should try to preserve the onboarding tests in the existing chrome suites with a reusable job, rather than creating another test. The existing suites would fail if that test would fail, so they're fairly dependent on onboarding passing. We could then create seperate sanity check for chrome, firefox, etc. using that same reusable job/workflow for the merge to master workflow
> * We might want to preserve the Firefox handholding checks that exist in case we want to expand the sanity testing; can we comment out or write a comment where we previously would skip a test on Firefox?

In reference to your concerns, I think a few of these will be addressed in the next PR. Trying to keep these CI PRs smaller for the sake of digestibility. I am just going through in chronological order of the master ticket Bruno made from our CI meeting.

DanielSinclair:
@BrodyHughes Do we still like all of these changes, or want to close?

BrodyHughes:
i'll close and look at it again next week and see how it looks with new CI changes we've made recently

PR #1677, opened 2024-08-22
swap: max swappable amount

greg-schrammel:
Fixes BX-1613
Figma link (if any):

increase the gas fee buffer, to prevent the issue in the ticket


walmat (approved):
🔥

PR #1680, opened 2024-08-26
chore: lavamoat upgrade for node v20

DanielSinclair:
Fixes BX-####
Figma link (if any):


brunobar79:
Revisiting if we need a mandatory bump of lavamoat

PR #1683, opened 2024-08-27
Help swaps flakiness

BrodyHughes:
Looks like a lot of changes but it's not really:

- wrapped the swaps flows in a `describe` wrapper, which makes the 'lines changes' really long, sorry
- added a few helpers to try and check exact transaction hashes by address to find the most recent sent txn by the test wallet, and see if the txn was successful. If not, it waits and checks again.
- this is better than some non-specific delays imo, and i think will help make the tests more consistent
- the two swaps flows were cut back several times, and so we were basically duplicating the entire test file to just swap USDC in the 2nd swap test. I instead just moved this to the end of the first test. This should make test times shorter, and hopefully help with flakiness by removing room for error


e2e/helpers.ts
```diff
@@ -776,6 +777,90 @@ export async function transactionStatus() {
...
+  maxAttempts = 20,
+): Promise<{
+  status: 'success' | 'failure' | 'timeout';
+  receipt: providers.TransactionReceipt | null;
+}> {
+  let attempts = 0;
+  while (attempts < maxAttempts) {
+    try {
+      console.log(`Checking transaction status, attempt ${attempts + 1}`);
+      const txnReceipt = await provider.getTransactionReceipt(txHash);
```
greg-schrammel, line 792: doesn't provider.waitForTransaction work here?

BrodyHughes: yeah lemme modify it a lil

BrodyHughes: hmm. i couldn't get it to work the way i tried. I don't mind the custom helper tbh if it works and includes some other useful logic. Do you see a big reason to use this method over the one i used?

brunobar79 (dismissed):
Conflicts need to be solved first

e2e/helpers.ts
greg-schrammel: not really just because it should do what your doing here, but if it didn't work that's fine

e2e/serial/swap/1_swapFlow1.test.ts
```diff
@@ -68,980 +59,1188 @@ const WALLET_TO_USE_ADDRESS = isFirefox
...
-  await findElementByTestIdAndClick({
-    id: 'swap-settings-flashbots-toggle',
-    driver,
-  });
-  await findElementByTestIdAndClick({
-    id: 'swap-settings-flashbots-toggle',
-    driver,
-  });
-});
+  // here
```
greg-schrammel, line 143: here

BrodyHughes: here

PR #1688, opened 2024-08-28
Fix pk back nav

BrodyHughes:
Fixes: BX-1619: 

previous behavior: 
1. navigate to wallet details
2. more menu
3. View PK
4. 'i've saved these words'
5. hit back nav
6. goes back to the pkey page instead of settings (incorrectly exposing PK)
(see ticket for repro video

FIxed behavior:
1-5 above
6. goes back to settings

POW:
<video src="https://github.com/user-attachments/assets/0152d67e-3a47-44c0-ac57-20529dcb9671"></video>

We had existing logic for navigating back to settings from this sort of state, just needed to pass it into the state nav


walmat (approved):
😘

derHowie (approved):
this is a classic, nice work

PR #1690, opened 2024-09-06
fix: prevent underflow error when parsing native asset fee and price

derHowie:
Fixes BX-1629

Changed how we parse native asset fee and price to prevent underflow errors thrown by bignumber.

```
Error: consolidatedTransactionsQueryFunction:  {
  "message": {
    "reason": "fractional component exceeds decimals",
    "code": "NUMERIC_FAULT",
    "fault": "underflow",
    "operation": "parseFixed"
  }
}
```

Screenshot 2024-09-06 at 7 28 06 AM

## What to test
Grunt.eth has transactions that were blowing up tx history. Reference the slack thread if you want to find more example addresses to test.


estebanmino (approved):
lgtm

src/core/utils/transactions.ts
```diff
@@ -278,19 +281,23 @@ export function parseTransaction({
...
     BigNumber.from(nativeAsset?.value || 0),
     nativeAsset?.asset.decimals || 0,
   );
 
-  const valueInNative = value.mulUnsafe(nativeAssetPrice).toString();
-
   const nativeAssetDecimals = 18; // we only support networks with 18 decimals native assets rn, backend will change when we support more
 
+  const nativeAssetPrice = FixedNumber.fromString(
+    typeof nativeAsset?.price === 'number'
```
greg-schrammel, line 293: in what case is it not a number? can we maybe assert this some other way?

derHowie: how would u suggest this be done

i just wanna check that it exists and still handle 0

greg-schrammel: wouldn't this work?
```ts
const nativeAssetPrice = FixedNumber.fromString(
    nativeAsset?.price ? nativeAsset.price.toFixed(nativeAssetDecimals).toString() : '0',
    nativeAssetDecimals,
  );
```

derHowie: This does look cleaner. I guess I prefer being more defensive since we're calling `toFixed` and this would blow up if we were accidentally sent a string.

greg-schrammel (approved):

PR #1691, opened 2024-09-06
fix: always bump user to tokens after a swap

derHowie:
Fixes BX-1620

From Daniel: "Small follow-up to the Perceived Finality PR. I think in all cases here we just bounce user to the Wallet Screen instead of Activity"

PoW: https://www.loom.com/share/714b35e54e2a49e09ae952fd0d482312

## What to test
Swap and make sure it takes you to the tokens screen


brunobar79 (approved):
LGTM!

PR #1702, opened 2024-09-16
update a lot of packages to fix ledger issues with chromedriver

BrodyHughes:
Fixes BX-####
Figma link (if any):


.github/workflows/build.yml
```diff
@@ -20,9 +20,15 @@ jobs:
...
-          cache: 'yarn'
-      - name: Install deps via Yarn
-        run: yarn setup
+          # cache: 'yarn'
+      - name: Clear cache, install deps via Yarn
+        run: |
+          echo "Clearing yarn cache..."
+          yarn cache clean
+          echo "Installing dependencies..."
+          yarn install --frozen-lockfile
```
brunobar79, line 29: Why?

.github/workflows/build.yml
```diff
@@ -48,6 +54,8 @@ jobs:
           name: node_modules.tar.gz
       - name: Unzip node_modules
         run: tar xzf node_modules.tar.gz
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
brunobar79, line 58: Also why?

brunobar79 (approved):
LGTM!

.github/actions/chromeTestsSetup/action.yml
```diff
@@ -12,4 +12,4 @@ runs:
       gh-access-token: ${{ inputs.gh-access-token }}
   - name: Install chrome
     shell: 'bash'
-    run: npx @puppeteer/browsers install chrome@121
```
derHowie, line 15: do we want to keep repinning this? wondering if we should try to be @latest again.

derHowie (changes requested):
Did you have any takeaways when you attempted to get those tests working? Any idea what's going on? Merging those skips into master makes me nervous because then we'd be able to add more bugs to those flows without knowing.

.github/actions/chromeTestsSetup/action.yml
BrodyHughes: - yeah I originally changed it to that but some other errors happened and bruno pinned it. i am honestly down for either but doing @latest does require we update chromedriver regularly.
- takeaways when trying to fix tests were I think we merged something in that broke it. the only thing currently broken is the actual executing swap. Here is the error i was seeing:
Screenshot 2024-09-17 at 1 34 18 AM
- my thought was since bx is moving slow and this release should get out we can unblock CI and then solve it afterwards

PR #1703, opened 2024-09-17
Fix BX-1641 & BX-1642

derHowie:
We were not properly clearing swap data when a user cleared the input or manually typed '0' into the swap inputs.

PoW: https://www.loom.com/share/5415e6ba392f474489728d7a1a50cbf1

## What to test
Reference the bugs detailed in the two attached tickets, but generally play the swap inputs and ensure that unnecessary market warnings are not displayed when clearing swap input data and that we don't see any wacky values displayed in the native currency display fields when swap input data is cleared from either input.


BrodyHughes (approved):
nice

PR #1708, opened 2024-09-23
fix: custom slippage

estebanmino:
Fixes BX-4828
Figma link (if any):

custom slippage wasn't working on prod so e2e was sending swaps with a very low slippage value

also edited e2e flow to correctly type values in slippage and swaps input to make e2e work


DanielSinclair (approved):

BrodyHughes (approved):
tested e2e and a swap with custom slippage

PR #1710, opened 2024-09-23
Ledger patch firefox friendly

brunobar79:
Fixes BX-####
Figma link (if any):

Patches cryptoassets dep from ledger to load the json files dynamically  using chrome's API instead of being imported directly in the bundle.  This makes the build to be at 3.3MB, which is below firefox limit of 4MB

Below you can see all the files being loaded correctly.
Screenshot 2024-09-23 at 5 57 05 PM

File size:
Screenshot 2024-09-23 at 5 59 28 PM

## What to test

- Ledger connection
- Ledger sending ERC20
- Ledger Swaps


BrodyHughes (approved):
Tested:
- Onboarding
- Swaps
- Sends

DanielSinclair (approved):

PR #1717, opened 2024-09-26
extend user native asset

estebanmino:
Fixes BX-1635
Figma link (if any):

using metadata external token query to get native assets for all chains as described in ticket

now we have two different hooks for this:

- useNativeAsset:  that will give you a ParsedAsset object, that contains the price of the asset for stuff like tx fees
- useUserNativeAsset: that will give you a ParsedUserAsset object, that contains the wallet's balance

### DIFF
--- a/src/core/resources/assets/externalToken.ts
+++ b/src/core/resources/assets/externalToken.ts
@@ -11,11 +11,7 @@ import {
   queryClient,
 } from '~/core/react-query';
 import { SupportedCurrencyKey } from '~/core/references';
-import {
-  AddressOrEth,
-  ParsedAsset,
-  ParsedUserAsset,
-} from '~/core/types/assets';
+import { AddressOrEth, ParsedAsset } from '~/core/types/assets';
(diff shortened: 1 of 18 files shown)


src/core/resources/assets/externalToken.ts
```diff
@@ -0,0 +1,152 @@
...
+  chainId,
+  currency,
+}: ExternalTokenArgs) {
+  await queryClient.prefetchQuery({
+    queryKey: externalTokenQueryKey({ address, chainId, currency }),
+    queryFn: externalTokenQueryFunction,
+    staleTime: EXTERNAL_TOKEN_STALE_TIME,
+    gcTime: EXTERNAL_TOKEN_CACHE_TIME,
+  });
+}
```
derHowie, line 132: do you think we'll end up using this prefetch?

derHowie (approved):
Code looks good

walmat (approved):
lgtm

PR #1720, opened 2024-09-30
addys summary for all native assets

estebanmino:
Fixes BX-1650
Figma link (if any):

before we only had 4 native assets to display when importing a new wallet with addys summary endpoint, because they were hardcoded

with this PR we'll start taking summary from summary_by_chain and I expanded useNativeAssets from data coming from backend so each time we add a new network this will be handled automatically without code changes

image


derHowie (approved):
nice, saw no issues

DanielSinclair (approved):

PR #1725, opened 2024-10-09
feat: add fee in native token on swap review sheet

derHowie:
Fixes BX-1606

We are showing the rainbow swap fee denominated in the native token rather than USD.

Screenshot 2024-10-09 at 12 10 01 PM
Screenshot 2024-10-09 at 12 11 45 PM
Screenshot 2024-10-09 at 12 23 14 PM

## What to test
Make sure the fee row looks fine and nothing broke.


DanielSinclair (approved):
Functionality looks good to me

PR #1733, opened 2024-10-16
fix transaction details "To" 

greg-schrammel:
Fixes BX-####
Figma link (if any):

fixes the "to" in activity details

previously "swapped" was displaing "to" as the address of the user 
Screenshot 2024-10-16 at 03 05 25

Screenshot 2024-10-16 at 03 00 55


src/core/utils/transactions.ts
```diff
@@ -175,20 +175,18 @@ const getAssetFromChanges = (
...
-  meta: PaginatedTransactionsApiResponse['meta'],
-  changes: {
-    direction: TransactionDirection;
-    asset: ParsedUserAsset;
-    address_to: Address;
-  }[],
-) => {
-  if (meta.type === 'approve') {
-    return meta?.approval_to;
+const getAddressTo = (tx: PaginatedTransactionsApiResponse) => {
```
benisgold, line 178: not sure what these `meta.type`s are, but should there be a case for `swap` instead of just defaulting?

derHowie (approved):

DanielSinclair (approved):

PR #1735, opened 2024-10-17
fix: include index when generating activity cell keys

derHowie:
We were using the transaction hash without appending an index when generating component keys and this was leading to overlaps in the UI.

Screenshot 2024-10-17 at 12 07 10 PM

## What to test
I think we were only getting punished for this key generation during specific race conditions. During the pending transaction flow you will sometimes see an overlap of transactions briefly and other times you will not. Check it out and definitely report back if you see the issue persist, but otherwise I wouldn't bang on this too hard.


src/entries/popup/hooks/useInfiniteTransactionList.ts
```diff
@@ -116,7 +116,7 @@ export const useInfiniteTransactionList = ({
         const txOrLabel = formattedTransactions[i];
         return typeof txOrLabel === 'string'
           ? txOrLabel
-          : txOrLabel.hash + txOrLabel.chainId;
+          : txOrLabel.hash + txOrLabel.chainId + i;
```
greg-schrammel, line 119: I think the solution is to dedupe them in `formattedTransactions` 
because `pendingTransactions` and `transactionsAfterCutoff` don't update together, they can add the same tx until `pendingTransactions` update
```ts
const formattedTransactions = useMemo(
    () =>
      Object.entries(
        selectTransactionsByDate([
          ...pendingTransactions,      // <---- this is adding the tx in pending
          ...transactionsAfterCutoff,  // <---- this is re-adding the same tx not pending anymore
        ]),
      ).flat(2),
    [pendingTransactions, transactionsAfterCutoff],
  );
```

derHowie: Yeah, this looks like a good change to make. But changing this key is very straight forward and will prevent the crash as a fallback. It will also allow us to have multiple rows for the same tx id if we ever need that flexibility.

greg-schrammel: my issue with adding the index is that the whole list has to be rerendered because all keys change
imagine the list 
```
- 0xb20...02-0 // <- this is pending 
- 0xE50...0e-1
- 0xab3...a3-2
```
when the first one is confirmed we update the list like this
```
- 0xb20...02-0 // <- same as the next one but confirmed
- 0xb20...02-1 // <- still the pending one
- 0xE50...0e-2 // <- same as the second in the last render, but with a new key now, so it's rerendered
- 0xab3...a3-3 // <- also is rerendered because new key
```
I think if we don't want to dedupe the `formattedTransactions` and show pending and confirmed as different rows, we could add the `tx.status` to the key

derHowie: Good point regarding rerenders. I've deduped them in the tx arrays. I think we should keep the key change just in case, but I'm now using status as you suggested.

greg-schrammel (approved):

DanielSinclair (approved):
Hasn't look this good and polished in so long. Thank you both 🥹

PR #1742, opened 2024-10-23
Bridge tx via Relay is classified as Swapping activity

derHowie:
Fixes BX-1470

The pending state of bridging currently says 'swapping', we want it to say bridging.

https://www.loom.com/share/5660617394154662b1c1e095a7560a61

## What to test
Perform a bridge, navigate to the activity screen, observe whether the pending state looks correct


src/entries/popup/pages/swap/index.tsx
```diff
@@ -605,6 +596,37 @@ export function Swap({ bridge = false }: { bridge?: boolean }) {
...
+      ];
+    const assetToBuyAddressToCompare =
+      assetToBuy?.[
+        assetToBuy?.chainId === ChainId.mainnet ? 'address' : 'mainnetAddress'
+      ];
+    return isLowerCaseMatch(
+      assetToSellAddressToCompare,
+      assetToBuyAddressToCompare,
+    );
+  }, [assetToBuy, assetToSell]);
```
greg-schrammel, line 612: wdyt of `isSameAssetInDiffChains(assetToSell, assetToBuy);`?

src/core/raps/references.ts
```diff
@@ -57,6 +57,7 @@ export interface RapSwapActionParameters<
   flashbots?: boolean;
   quote: QuoteTypeMap[T];
   address?: Address;
+  isBridge?: boolean;
```
greg-schrammel, line 60: I think since `isBridge` doesn't change the function behavior, only metadata, it can be confusing having it in the parameters
can we derive when needed with `isSameAssetInDiffChains(assetToSell, assetToBuy)` or something?

DanielSinclair (approved):
Perfect 👌

src/entries/popup/pages/swap/index.tsx
derHowie: 👍

src/core/raps/references.ts
derHowie: 👍

greg-schrammel (approved):

PR #1748, opened 2024-10-26
add isHardwareWallet to swap analytics event

greg-schrammel:
Fixes BX-####
Figma link (if any):

tracks if swap/bridge signer is a hardware wallet in amplitude


DanielSinclair (approved):
Pushed up a tiny tweak to drop the `is` and add the param to the bridge sister event, but otherwise worked as expected

BrodyHughes (approved):

PR #1753, opened 2024-11-01
Copy checksummed address

JacobHomanics:
When a user copies a token address to their clipboard, the copied address should be the proper checksummed version.
This is done by wrapping the address that is copied inside a `getAddress()` call from viem.

Screenshot 2024-11-01 at 12 58 40 AM
Screenshot 2024-11-01 at 12 58 51 AM

## What to test
Test if the copied addresses of tokens are properly checksummed. 

## Notes
This is my first time formally contributing to the codebase. I am not familiar with it and am not sure how to set up a local environment. Thus I just made the proposed changes to the best of my ability/knowledge. Glad to work through the problem to get the feature where it needs to be.

It is possible that this feature should be expanded to ANY address, rather than just token addresses.

Cheers!

### DIFF
--- a/src/core/utils/copy.ts
+++ b/src/core/utils/copy.ts
@@ -1,4 +1,5 @@
 import { AddressZero } from '@ethersproject/constants';
+import { getAddress } from 'viem';
 
 import { triggerToast } from '~/entries/popup/components/Toast/Toast';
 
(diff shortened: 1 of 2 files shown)


DanielSinclair:
@JacobHomanics Thanks for spotting this! I think we have a few other areas where we have inconsistencies in how we display / copy addresses, so we'll build off of your PR

JacobHomanics:
> @JacobHomanics Thanks for spotting this! I think we have a few other areas where we have inconsistencies in how we display / copy addresses, so we'll build off of your PR

great!

Glad to help out if I can be of any use! Looking forward to this improvement

greg-schrammel (approved):
nice!

DanielSinclair (approved):

PR #1762, opened 2024-11-07
wallet telemetry alignment

DanielSinclair:
Fixes BX-1694
Figma link (if any):

- alignment with App telemetry
- setting up Sentry and Rudderstack identify in new `useEffect` to absorb wallet switches
- added `getWalletContext` util to prepare hashes and wallet types consistently throughout the codebase
- added `walletContext` overrides in track and screen to support scenarios where the actioned wallet is different than the global selected wallet
- moved the internals of the `useEffect` in `App.tsx` for analytics setup in `TelemetryIdentifier` component that lives inside the Auth provider
- overrode wallet context in `dappPromptConnectApproved`, `dappPromptConnectRejected`, `dappPromptSendTransactionApproved`, `dappPromptSendTransactionRejected`, `dappPromptSignMessageApproved`, `dappPromptSignTypedDataApproved`, `dappPromptSignMessageRejected`, `dappPromptSignTypedDataRejected` where the user can switch wallets prior to signing; this should cover all edge cases

## What to test
- confirmed event flow on rudderstack ✅ 
- no race condition with Sentry initialization or logs fired by `Sentry.setUser` if useEffects are called in unexpected order


DanielSinclair:
@derHowie @BrodyHughes When you get a sec to review, would like to include in next release

derHowie (approved):

BrodyHughes (approved):
Set up rudderstack dev env and then tested switching wallets, connecting to a dapp, confirming and rejecting a txn and most things populated as expected in rudderstack. DM'd Daniel about the one area i wasn't seeing populate.

DanielSinclair:
@BrodyHughes Believe these events are working as expected now 
Screenshot 2024-11-26 at 3 53 38 AM

DanielSinclair:
Did an extensive refactor here to move the internals of our analytics setup `useEffect` in `App.tsx` to a `TelemetryIdentifier` component that relys on the `AuthProvider` context and `useAuth`. The issue we encountered was that the `getWallet` call in `getWalletContext` would error while the keychain was locked and halt the identification; this occured on each first-load as the extension launches with a locked keychain. We now do an identify before the keychain is loaded (for the device id and hash), and after the keychain is loaded (with the walletType and above). This also occurs on wallet switch as you would expect.

There is a pattern of issues that I separately identified here that I'll work on in a follow-up. On launch, we trigger a `popupOpened` event. We also trigger a series of `screen()` events as we route to the lock screen, home screen, etc. These events are dropped because they often fire before the `deviceId` is rehydrated from storage. Likewise, they may fire without a hash in the walletContext, or there will be a mismatch between the current identification in the analytics client and the newly selected wallet, so events could be attributed to the wrong context. We would need to fix this with a lock and queue of some sort, similarly to how the background event queue works.

derHowie (approved):
👌

PR #1768, opened 2024-11-13
Match waitForNodeAck logic to be same as in app

jinchung:
* Surrounded the `waitForNodeAck` logic in try catch and follow the same retry pattern up to 10 max retries

Brody had confirmed PoW on Zora.


brunobar79 (approved):
👍 👍 👍

PR #1778, opened 2024-12-04
allowlist advisory for @solana/web3.js

BrodyHughes:
I tried a package resolution (@solana/web3.js 1.95.8) but it's not effective because the advisory marks all versions (>=0) as vulnerable. Adding this to the allowlist is necessary since the advisory considers every version compromised, even though 1.95.8 contains the actual fix. 

This package is just a transitive dependency through @trezor/connect as seen here:

```
npm ls @solana/web3.js
browser-extension@1.5.64 /Users/brodyhughes/browser-extension-master
└─┬ @trezor/connect-web@9.4.2
  └─┬ @trezor/connect@9.4.2
    ├─┬ @trezor/blockchain-link-types@1.2.1
    │ └── @solana/web3.js@1.95.8
    └─┬ @trezor/blockchain-link@2.3.1
      ├── @solana/web3.js@1.95.8 deduped
      └─┬ @trezor/blockchain-link-utils@1.2.1
        └── @solana/web3.js@1.95.8 deduped
```


brunobar79:
Adding some context here:

1) Rainbow is def. *NOT* compromised.
2) We don't support solana or use this library directly, it's a dependency from trezor's SDK
3) We have [version 1.90.2](https://github.com/rainbow-me/browser-extension/blob/master/yarn.lock#L4515) while the compromised versions were 1.95.6 and 1.95.7
4) We'll need to patch this in the future to lock it onto 1.95.8 so we can get rid of the CI warnings, which at the moment seem to be incorrect

PR #1781, opened 2024-12-11
Replace broken Edge browser logo with working version from jsDelivr CDN

savvar9991:
Update Edge browser logo URL
Changed the Edge browser logo source from Wikimedia to jsDelivr CDN for better reliability and performance.The previous icon was not displayed correctly
Before:
image
After:
image


DanielSinclair (approved):
Thanks!

PR #1782, opened 2024-12-11
More precisely default to -1 for currentNonce when it does not exist

jinchung:
This change sets the default "currentNonce" to -1 to distinguish between an actual current nonce of 0 vs an unknown current nonce (e.g, for a fresh wallet we just created that has no outgoing txns, or a freshly imported wallet that we hadn't previously tracked.)

This helps us treat nonces vs txn counts accurately.


derHowie (approved):

BrodyHughes (approved):
successful nonce 0 txn: https://etherscan.io/tx/0x05a5b75d7c865c455694c0b91cdd607f2ae2cb1cd3d93b85cc025937bab72435

PR #1797, opened 2025-01-03
Surface UUID in a more user friendly way for support

BrodyHughes:
Fixes BX-1717

Goal as explained in the ticket:

- Add diagnostic Cmd K actions when when searching Diagnostics (should be hidden otherwise)
- Create a Copy UUID action that copies the deviceId to the clipboard with a toast
- When clicking the version number at the bottom of the Settings, trigger the same Copy UUID action

- Added a hook for handling user UUID and the copying action / toast
- Added the commandK action searchable by 'Diagnostics' as described in ticket. if 'Diagnostics' in it's entirety isn't searched for, the command will not surface
- Also added the ability to copy the UUID to clipboard by double-clicking the version number on the settings sheet

<video src="https://github.com/user-attachments/assets/9edfa6ea-7837-4e01-8257-7e59f50b8af6" />

## What to test

- is the command hidden until 'diagnostics' is searched for in its entirety?
- does it copy the UUID to the clipboard?

### DIFF
--- a/src/entries/popup/components/CommandK/useCommands.tsx
+++ b/src/entries/popup/components/CommandK/useCommands.tsx
@@ -174,13 +174,6 @@ export const getStaticCommandInfo = (): CommandInfo => {
       symbolSize: 15,
       type: SearchItemType.Shortcut,
     },
-    copyAppUUID: {
-      name: getCommandName('diagnostics'),
-      page: PAGES.HOME,
-      symbol: 'square.on.square',
-      symbolSize: 15,
-      type: SearchItemType.Shortcut,
-    },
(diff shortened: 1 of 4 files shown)


DanielSinclair (approved):
Added a `diagnostics` search tag so we can add a few more diagnostics in the future, and reorganized a bit with a pseudo page section. Functionality lgtm!

DanielSinclair:
I'll try to remember to prevent translations for these strings so that the menu can always be found no matter the user's language

brunobar79 (approved):
🔥 🔥 🔥

PR #1799, opened 2025-01-10
[Analytics] Replace `_execute_action` with Custom Command

BrodyHughes:
Fixes BX-1625

Currently our extension uses `_execute_action` for the Alt+Shift+R shortcut to open the popup. While functional, this reserved command doesn't dispatch any events we can hook into for analytics. [From documentation](https://developer.chrome.com/docs/extensions/reference/api/commands#action_commands):

```
The _execute_action (Manifest V3), _execute_browser_action (Manifest V2), and _execute_page_action
(Manifest V2) commands are reserved for the action of trigger your action, browser action, or page action 
respectively. These commands don't dispatch [command.onCommand] events like standard commands.
```

This PR replaces `_execute_action` with a standard command that checks `command.onCommand` for the execution of this shortcut, and then runs `chrome.action.openPopup()` in our `background.js` service-worker script. This gives us identical popup opening behavior while allowing us to track shortcut usage through the `command.onCommand` listener.

The change does not affect users at all; the shortcut remains the same and popup opening performance is unchanged since we're using Chrome's native popup API.

## What to test

- Does extension opening shortcut work identical to how it did before?
- Does analytics for it populate?

### DIFF
--- a/e2e/serial/swap/1_swapFlow1.test.ts
+++ b/e2e/serial/swap/1_swapFlow1.test.ts
@@ -174,7 +174,7 @@ it('should be able to interact with slippage settings', async () => {
   await typeOnTextInput({
     id: 'slippage-input-mask',
     driver,
-    text: '5',
+    text: '6',
   });
   const warning = await findElementByTestId({
     id: 'swap-settings-slippage-warning',
--- a/src/entries/popup/pages/swap/SwapSettings/SwapSettings.tsx
+++ b/src/entries/popup/pages/swap/SwapSettings/SwapSettings.tsx
@@ -263,7 +263,7 @@ export const SwapSettings = ({
   }, [onDone, setSettings, slippage, source]);
 
   const slippageWarning = useMemo(
-    () => (Number(slippage) >= 3 ? 'loss' : undefined),
+    () => (Number(slippage) >= 6 ? 'loss' : undefined),
     [slippage],
   );
 


walmat (approved):
Any performance implications to always listening to chrome commands? Also, any need for other browser support?

BrodyHughes:
> Any performance implications to always listening to chrome commands? Also, any need for other browser support?

i dont think it should have any performance implications since it's event-driven and only runs when a keyboard shortcut is pressed. I did assume firefox would work with this but i dont think it does. ill check / fix that today.

walmat (approved):

src/entries/background/index.ts
```diff
@@ -18,6 +20,26 @@ import { handleTabAndWindowUpdates } from './handlers/handleTabAndWindowUpdates'
 import { handleWallets } from './handlers/handleWallets';
 require('../../core/utils/lockdown');
 
+declare const browser: Browser;
+
+const browserType = process.env.BROWSER;
```
DanielSinclair, line 25: This might not work in a prod build; I think we only use this for e2e/local builds

DanielSinclair: We have a `getBrowser` util here: https://github.com/rainbow-me/browser-extension/blob/97b914a299f1454b30b6ffbd87080929a2c1f097/src/entries/popup/hooks/useBrowser.ts#L8C17-L8C27

src/entries/background/index.ts
```diff
@@ -40,3 +62,10 @@ popupMessenger.reply('rainbow_updateWagmiClient', async () => {
   const { rainbowChains } = getRainbowChains();
   updateWagmiConfig(rainbowChains);
 });
+
+browserAPI.commands.onCommand.addListener((command: string) => {
```
DanielSinclair, line 66: Could be crazy, but thought `chrome` mapped one to one to `browser`, so thought `chrome.commands` should work in Firefox. If so, I think we could reduce the need for the polyfill and just do a check in the `openPopup` wrapper function for whichever of `browserAction` or `action` is defined and call the appropriate one

src/analytics/event.ts
```diff
@@ -137,6 +137,10 @@ export const event = {
    * Called when the dapps hits the rate limit per second or minute.
    */
   dappProviderRateLimit: 'dapp.provider.rate_limit',
+  /**
+   * Tracks when the extension popup is opened via keyboard shortcut (Alt+Shift+R)
+   */
+  extensionOpenViaShortcut: 'extension.open.shortcut',
```
DanielSinclair, line 143: Nitpicky, but might rename to `browserShortcutTriggered` with a `action` type (i.e. `launch`) to align with the other `keyboardShortcutTriggered` and `keyboardNavigationTriggered` events. We may add other actions in the future

src/entries/background/index.ts
BrodyHughes: okay i think youre right actually.

src/analytics/event.ts
BrodyHughes: lemme know if that works.

brunobar79 (approved):

PR #1809, opened 2025-01-20
fix: updated Discord links and replaced Twitter with X

NikolaiKryshnev:


DanielSinclair:
The Discord links are actually just for mock data that we use for testing and aren't visible for end users.

PR #1811, opened 2025-01-21
Add Chrome flags to fix CI test startup

BrodyHughes:
Chrome wasn't starting properly in CI last week. Added a couple flags to our chrome args seemed to fix it.

Added `--no-sandbox` and `--disable-dev-shm-usage` flags to Chrome test initialization to fix the "session not created" errors in CI. These flags are required for Chrome to run properly in containerized CI environments where the default sandbox permissions and shared memory settings are insufficient.

More about the flags:
- `--no-sandbox`: Disables Chrome's sandbox security feature which often can't get the necessary permissions in CI containers
- `--disable-dev-shm-usage`: Forces Chrome to use /tmp instead of /dev/shm for shared memory, preventing crashes from insufficient memory allocation in CI containers
- `--disable-extensions-except=build/`: Only leaves our extension enabled. Helps future proof tests a lil.

The error we were seeing:
```
SessionNotCreatedError: session not created: Chrome failed to start: exited normally.
(session not created: DevToolsActivePort file doesn't exist)
```

### DIFF
--- a/e2e/helpers.ts
+++ b/e2e/helpers.ts
@@ -115,20 +115,11 @@ export async function initDriverWithOptions(opts: {
   let driver;
   const args = [
     'load-extension=build/',
-    // '--auto-open-devtools-for-tabs',
     '--log-level=0',
     '--enable-logging',
-    // test
     '--no-sandbox',
     '--disable-dev-shm-usage',
-    '--disable-background-networking',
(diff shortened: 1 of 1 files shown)


DanielSinclair:
Thank you!

walmat (approved):

PR #1813, opened 2025-01-22
Consume suppported networks on runtime instead of build time

walmat:
Fixes BX-1740, BX-1675

Version bump Zustand to `4.5.5` to match the app and is necessary in order to bring over `createQueryStore`. Adds in `createQueryStore` which solely powers backendNetworks as of now. Also replaces the `references/chains.ts` file with consumption of the backendNetworks store instead. 

Other changes:
- `wagmiConfig` initialization is now powered by backendNetworks
- replaced `userChains` -> `getSupportedChains` with backendNetworks
- replaced local chain assets (minus custom assets) with the backend network version from Cloudinary

Things left to do:
- [ ] Figure how to replace [chainIdMap](https://github.com/rainbow-me/browser-extension/blob/0169120d3a40052f703ec2664ba64416adfe3c0a/src/core/utils/userChains.ts#L33) with backendNetworks
- [ ] Figure out how to handle updating the userChains / rainbowChains stores when the backendNetworks store changes data
- [x] https://linear.app/rainbow/issue/BACK-1595/add-badges-for-bx

// will do later once out of draft

## What to test

// will do later once out of draft


src/core/state/backendNetworks/backendNetworks.ts
```diff
@@ -18,6 +18,9 @@ import {
 } from '~/core/types/chains';
 import { GasSpeed } from '~/core/types/gas';
 
+import { useRainbowChainsStore } from '../rainbowChains';
+import { useUserChainsStore } from '../userChains';
```
DanielSinclair, line 22: I think we may want to use `userChainsStore` directly (and similar for the other hooks) because `/core` components can be consumed outside of the React tree in the other worker script entrypoints. Not sure if Zustand or the compiler would automatically handle that when initializing the hook and calling `getState()`

src/core/state/backendNetworks/backendNetworks.ts
```diff
@@ -0,0 +1,595 @@
...
+    //   return 'sanko';
+
+    case ChainId.scroll:
+      return 'scroll';
+    case ChainId.zksync:
+      return 'zksync-era';
+    case ChainId.zora:
+      return 'zora';
+    case ChainId.linea:
+      return 'linea';
```
DanielSinclair, line 590: In the [NFT query](https://github.com/rainbow-me/browser-extension/blob/af6b2daf5b88a8bc478dd7a6f66405a5a4e3d2b3/src/core/resources/nfts/galleryNfts.ts#L55) we fetch all NFTs for active networks and `gnosis` (for POAPs), even when they are not officially supported networks. That's the main reason we have had to maintain chain naming mappings (to stay in sync with SimpleHash) and append to `ChainName` and `ChainId` types. Would love to just get this as a value from backend so we can blow away all of that logic and types. We also don't use the `nftProxy` right now because Chris did work to optimize pagination for direct fetching from SimpleHash; that's probably why this data need diverged from App.

src/core/state/backendNetworks/backendNetworks.ts
```diff
@@ -150,6 +153,78 @@ function createParameterizedSelector<T, Args extends unknown[]>(
...
+ */
+const handleUpdateUserChains = (networks: BackendNetwork[]) => {
+  const { userChains, addUserChain } = useUserChainsStore.getState();
+
+  const newChains = networks.filter((network) => !userChains[+network.id]);
+  // if there are no new chains, then we don't need to update anything
+  if (!newChains.length) return;
+
+  newChains.forEach((chain) => addUserChain({ chainId: +chain.id }));
+};
```
DanielSinclair, line 171: Have thought `userChains` and `addUserChain` are a bit over-complicated for that store, and require us to do multiple state updates across both stores each time anything changes. I wonder if we can just get away with a synthetic `useChains` that stitches `rainbowChains` and `userChains` together, rather than maintaining the list in both stores. `userChains` essentially is just toggling the network on/off and ordering, but we could accomplish that by only keeping record of disabled networks and ordering with a `1 -> 137 -> 8453` linked list that doesn't need to understand the full chain list (since every other network can just be appended to the end in `useChains`). 

Screenshot 2025-01-24 at 6 36 23 PM

src/core/state/backendNetworks/backendNetworks.ts
```diff
@@ -150,6 +153,78 @@ function createParameterizedSelector<T, Args extends unknown[]>(
...
+        rpcUrl: chain.rpcUrls.default.http[0],
+      });
+      return;
+    }
+
+    // if we have an existing chain, then we update the existing chain
+    // and set the active RPC url to the backend network provided RPC url
+    updateCustomRPC({ chain });
+    setActiveRPC({ chainId: +chain.id, rpcUrl: chain.rpcUrls.default.http[0] });
+    return;
```
DanielSinclair, line 224: I think we'd only want to call `setActiveRPC` to update a custom chain that becomes a supported chain (old public RPC -> Rainbow proxy RPC). This currently catches the scenario where a user has previously switched the default RPC for a supported chain from Rainbow to i.e. Flashbots, so we wouldn't want to override that preference

src/core/state/backendNetworks/backendNetworks.ts
```diff
@@ -160,6 +235,12 @@ export const useBackendNetworksStore = createQueryStore<
...
       set((state) => {
         if (isEqual(state.backendNetworks, data)) return state;
+
+        void Promise.all([
+          handleUpdateUserChains(data.networks),
+          handleDiffAndMergeRainbowChains(data.networks),
+        ]);
+
         return {
           backendChains: transformBackendNetworksToChains(data.networks),
```
DanielSinclair, line 245: Could be a bottleneck, but I like the idea of only storing `backendNetworks` so that we can do transform at boot-time. Wagmi/Viem have had a few breaking changes with their `Chain` type that I can remember, so we'd need to catch those type differences before hydrating `backendChains`, and then do a migration to re-transform in the state store. Whereas "dumb" backend caching here inherently requires type validity upon hydration

src/core/state/userChains/index.ts
```diff
@@ -52,10 +52,10 @@ const initialChains = () => {
...
-  const supportedMainnetChainIds = useBackendNetworksStore
+  const sortedSupportedChainIds = useBackendNetworksStore
     .getState()
-    .getSupportedMainnetChainIds();
-  return supportedMainnetChainIds;
+    .getSortedSupportedChainIds();
+  return sortedSupportedChainIds;
 };
 
 export const userChainsStore = createStore<UserChainsState>(
```
DanielSinclair, line 61: We should be able to remove `migrations` here since `handleUpdateUserChains` is handling the diffing.

src/core/state/rainbowChains/index.ts
```diff
@@ -20,7 +22,7 @@ export interface RainbowChain {
 
 export interface RainbowChainsState {
```
DanielSinclair, line 23: We should be able to remove migrations here since `handleDiffAndMergeRainbowChains` is handling the diffing and we actually haven't had a breaking change (only appended networks since beginning of history) to me knowledge.

src/core/state/rainbowChains/index.ts
```diff
@@ -1,17 +1,19 @@
...
   mergeNewOfficiallySupportedChainsState,
   removeCustomRPC,
   replaceChainsWithInitial,
-} from '~/core/state/rainbowChains/utils';
-import { ChainId } from '~/core/types/chains';
-import { persistOptions } from '~/core/utils/persistOptions';
+} from './utils';
 
 export interface RainbowChain {
   activeRpcUrl: string;
```
DanielSinclair, line 19: I'm not exactly sure how we ended up with sub-chains here like you mentioned in your message, as opposed to an RPC list. That's a headache. I could foresee blowing this up and bundling the useful components of `rainbowChains` and `userChains` into a `chainPreferences` store that only handles preference updates, like network disabled, network ordering, custom RPC list, and RPC selection — while `backendChains` remains immutable with these preferences applied on top. We would need to keep in mind that state still exists in `rainbowChains` and `userChains` so would need check if they're available and do a one-time migrate and delete.

walmat:
Closing in favor of new networkStore infrastructure. Will reference this to pull in some of the functionality, but most of this is stale now.

PR #1829, opened 2025-01-31
Add unified networks store

walmat:
Fixes BX-####
Figma link (if any):


package.json
```diff
@@ -157,6 +157,7 @@
     "react": "18.2.0",
     "react-beautiful-dnd": "13.1.1",
     "react-dom": "18.2.0",
+    "react-fast-compare": "3.2.2",
```
walmat, line 160: can remove this since we aren't going to use it for comparisons
```suggestion
```

src/core/references/chains.ts
walmat, line 1: this file will be deleted once integration begins, changes here are only to keep the linter from complaining

src/core/resources/networks/networks.ts
walmat, line 1: the networks fetch now fetches both backend driven and custom backend driven networks in parallel

package.json
```diff
@@ -61,6 +61,7 @@
     "anvil:optimism": "anvil --fork-url $(./scripts/get-rpc-url.sh optimism)",
     "anvil:kill": "lsof -i :8545|tail -n +2|awk '{print $2}'|xargs -r kill -s SIGINT",
     "test": "./scripts/unit-tests.sh",
+    "test:file": "vitest",
```
walmat, line 64:
```suggestion
```

PR #1836, opened 2025-02-07
Cleanup hardcoded network values and replace with backend driven

walmat:
Fixes BX-1738


src/core/resources/_selectors/assets.ts
```diff
@@ -82,24 +71,6 @@ export function selectUserAssetsListByChainId(assets: ParsedAssetsDictByChain) {
...
-    [ChainId.bsc]: mapAddresses(assets[ChainId.bsc]) || [],
-    [ChainId.polygon]: mapAddresses(assets[ChainId.polygon]) || [],
-    [ChainId.arbitrum]: mapAddresses(assets[ChainId.arbitrum]) || [],
-    [ChainId.base]: mapAddresses(assets[ChainId.base]) || [],
-    [ChainId.zora]: mapAddresses(assets[ChainId.zora]) || [],
-    [ChainId.avalanche]: mapAddresses(assets[ChainId.avalanche]) || [],
-    [ChainId.ink]: mapAddresses(assets[ChainId.ink]) || [],
-  };
-}
-
```
walmat, line 102: this was unused, so removed it

src/core/state/networks/networks.ts
```diff
@@ -393,6 +395,28 @@ export const networkStore = createQueryStore<
...
+      return networks.backendNetworks.networks.reduce((acc, curr) => {
+        const mainnetId = +curr.mainnetId;
+        if (!acc[mainnetId]) {
+          acc[mainnetId] = [];
+        }
+        acc[mainnetId].push(curr);
+        return acc;
+      }, {} as Record<number, BackendNetwork[]>);
+    }),
+
```
walmat, line 419: this replaces the `chainIdMap`, which we can use to construct the `chainsLabelMap`

src/entries/popup/components/SideChainExplainer.tsx
```diff
@@ -22,6 +22,8 @@ export const getSideChainExplainerParams = (
   onDismiss: VoidFunction,
 ) => {
   const chainName = chainIdToNameMapping[chainId];
+
+  // FIXME: https://linear.app/rainbow/issue/BACK-1452/provide-locale-copy-from-the-backendnetworks-for-explainers-so-that-we
```
walmat, line 26: fyi this has been on the backend driven todo list for a bit. Not sure if we have a good solution yet, but will raise in stand up on Monday. cc @fringlesinthestreet

PR #1837, opened 2025-02-08
Implement new token search improvements

walmat:
Fixes BX-####
Figma link (if any):


brunobar79:
@walmat  do we still want this in?

PR #1845, opened 2025-02-19
[Analytics] token favorites / hiding analytics events

BrodyHughes:
- adds analytics tracking for favorite / unfavorite & hide / unhide

Favorite / unfavorite POW:

<Video src="https://github.com/user-attachments/assets/2cd54d8c-04a1-49b3-b07c-b39e37ad273d"></Video>

Hide / unhide POW:
<Video src="https://github.com/user-attachments/assets/7df5a96c-50bd-475d-b859-04eb665c74dd"></Video>

## What to test
- do analytics work? 
- are we reporting what we want?


walmat:
@BrodyHughes can we wait on this one until my networks stuff goes in? They both touch this file and will need a rebase.

brunobar79:
@BrodyHughes  needs to rebase after #1847

src/core/state/favorites/index.ts
```diff
@@ -37,32 +38,67 @@ export interface FavoritesState {
...
+      isAdding: boolean,
+    ) => {
+      const updatedState = get();
+      const updatedFavorites = updatedState.favorites[chainId] || [];
+      const afterLength = updatedFavorites.length;
+
+      const analyticsData = {
+        token: {
+          address,
+          chainId,
```
derHowie, line 55: Since we're getting favorites and safely accessing favorites by chainId within `addFavorite` and `removeFavorite`, we are kind of duplicating that logic here.

If you pass the edited favorites array (or its length) to this function we could reduce some repetition and then would also be able to remove this from the store's setup completely.

src/core/state/hiddenAssets/hiddenAssets.ts
```diff
@@ -26,15 +27,40 @@ export const hiddenAssetsStore = createStore<HiddenAssetState>(
...
+      const isNowHidden = updatedState.hidden[address]?.[uniqueId] || false;
+      // Calculate total hidden assets for this address
+      const totalHiddenForAddress = Object.values(
+        updatedState.hidden[address] || {},
+      ).filter((isHidden) => isHidden).length;
+
+      const assetHiddenAnalytics = {
+        token: {
+          address: uniqueId.split('-')[0], // Extract the asset address from uniqueId
+          chainId: parseInt(uniqueId.split('-')[1]), // Extract the chainId from uniqueId
```
derHowie, line 51: You could grab these both in the same .split call above.

src/core/state/hiddenAssets/hiddenAssets.ts
```diff
@@ -26,15 +27,40 @@ export const hiddenAssetsStore = createStore<HiddenAssetState>(
...
+        hiddenAssets: {
+          totalHidden: totalHiddenForAddress,
+        },
+      };
+      // Track the appropriate event based on whether we hid or unhid
+      if (isNowHidden) {
+        analytics.track(analytics.event.assetHidden, assetHiddenAnalytics);
+      } else {
+        analytics.track(analytics.event.assetUnhidden, assetHiddenAnalytics);
+      }
```
derHowie, line 63:
```suggestion
      const event = isNowHidden ? analytics.event.assetHidden : analytics.event.assetUnhidden;
      analytics.track(event, assetHiddenAnalytics);
```

derHowie (commented):
Looks good! Just some small nits.

src/core/state/favorites/index.ts
```diff
@@ -37,32 +38,65 @@ export interface FavoritesState {
...
-    setFavorites: (favorites) => set({ favorites }),
-    addFavorite: ({ address, chainId }: UpdateFavoritesArgs) => {
-      const { favorites } = get();
-      const currentFavorites = favorites[chainId] || [];
-      set({
-        favorites: {
-          ...favorites,
-          [chainId]: [...currentFavorites, address],
+  (set, get) => {
+    const trackFavoriteChange = (
```
walmat, line 42: is there any reason to put this function inside of the Zustand store?

src/core/state/hiddenAssets/hiddenAssets.ts
```diff
@@ -26,15 +27,39 @@ export const hiddenAssetsStore = createStore<HiddenAssetState>(
...
+        updatedState.hidden[address] || {},
+      ).filter((isHidden) => isHidden).length;
+
+      const [assetAddress, chainIdStr] = uniqueId.split('-');
+
+      const assetHiddenAnalytics = {
+        token: {
+          address: assetAddress,
+          chainId: parseInt(chainIdStr),
+          walletAddress: address,
```
walmat, line 52: idk if this is still the case, but are we sure we want to pass this data across for analytics? IIRC wallet addresses were treated as sensitive data, and this would expose a lot of tracking that could be deemed as not so savory.

BrodyHughes: yeah probably unnecessary tbh. i can remove.

src/core/state/favorites/index.ts
BrodyHughes: It was my understanding that it doesn't negatively impact the store and i thought i was better to centralize the logic in one place than duplicating the logic wherever we are importing the `useFavoritesStore`

walmat: We could keep it in this file, but I'd move it outside of the store creator. Tbh might just be a nitpick from me but seems like a weird pattern.

BrodyHughes: na its a valid nit. can move.

walmat (approved):

BrodyHughes:
need to fix e2e

DanielSinclair (approved):

PR #1847, opened 2025-02-21
Fix: Network data loss during internal build upgrade testing

walmat:
Fixes BX-1766

- Fixed `sortNetworks` not being called anymore in the networks page
- `useConfig` from wagmi wasn't detecting chain changes from the `WagmiConfigUpdater`, so consuming `wagmiConfig` fixes it directly
- Fixes adding a custom chain which re-enabled all chains

https://github.com/user-attachments/assets/018b29ac-d2e7-4211-8ca8-9a41b9ff9de0

## What to test
migration patterns as mentioned in the Linear ticket


e2e/helpers.ts
```diff
@@ -200,6 +204,8 @@ export async function getExtensionIdByName(
       .getText();
     return text;
   } else {
+    return 'lccajkdjonnokmddpnclgcihidkeklij';
```
DanielSinclair, line 207: I think this would need to be dynamic; each install gets a random id IIRC

walmat: Yes the method to find the extension by Id is broken in chrome 133 so I had this hardcoded to test. I'll revert it was mistakenly committed.

BrodyHughes (approved):
1.9.86 > 1.9.91 upgrade path now migrates networks as expected! also tested swaps 👍

BrodyHughes (approved):
migrated from 1.5.86 > this branch ten times with no issues.

derHowie (approved):

PR #1849, opened 2025-03-03
Bump provider to 0.1.2

BrodyHughes:
- Bumped provider to version 0.1.2
- Same bump happening on app side

## What to test
- Swapping


BrodyHughes:
- Tested several swaps across different chains on uniswap. 
- Tested dapp interactions on our test dapp shown in above video.
- Switched chains in extension and on the dapp and saw them reflected correctly on both sides.

PR #1852, opened 2025-03-07
increase default slippage warning from 5% > 6%

BrodyHughes:
- our default slippage for L2s is 5%, but 5% throws a warning for slippage being high. This adjusts it to 6%, so users don't see the warning now unless they manually increase slippage.

Before with default slippage:
Screenshot 2025-03-07 at 11 34 14 AM

After with default slippage:
Screenshot 2025-03-07 at 11 43 57 AM


derHowie:
Can you clean up the commit history on this? 6 commits for a 1-liner 😅

walmat:
> Can you clean up the commit history on this? 6 commits for a 1-liner 😅

does it matter? it gets squashed into one when we merge anyway, right?

derHowie (approved):

PR #1858, opened 2025-03-21
Use Target Address From Swap

derHowie:
bumped swaps, deferring to a now variable set of swap target addresses

## What to test
Typical swap flow with the addition of swaps including low liquidity RSTs (small buys after launch, then dumping into low liquidity pool), and RST for RST.

### DIFF
--- a/src/core/raps/utils.ts
+++ b/src/core/raps/utils.ts
@@ -95,7 +95,10 @@ const getStateDiff = async (
 ): Promise<unknown> => {
   const tokenAddress = quote.sellTokenAddress;
   const fromAddr = quote.from;
-  const toAddr = getTargetAddressForQuote(quote);
+  const toAddr =
+    quote.swapType === 'normal'
+      ? getTargetAddressForQuote(quote)
+      : (quote as CrosschainQuote).allowanceTarget;
   const tokenContract = new Contract(tokenAddress, erc20Abi, provider);
 
   const { number: blockNumber } = await (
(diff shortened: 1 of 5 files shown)


src/core/raps/utils.ts
```diff
@@ -95,22 +95,27 @@ const getStateDiff = async (
 ): Promise<unknown> => {
   const tokenAddress = quote.sellTokenAddress;
   const fromAddr = quote.from;
-  const { chainId } = await provider.getNetwork();
-  const toAddr =
-    quote.swapType === 'normal'
-      ? getRainbowRouterContractAddress(chainId)
```
jinchung, line 101: we need to update this on the app as well - it is not technically wrong (the `allowanceTarget` and the `to` are the same for crosschain swaps), but for the sake of consistency can you keep the CrosschainQuote flow with the allowance target the same, and just replace the `getRainbowRouterContractAddress` with `getTargetAddressForQuote`

BrodyHughes (approved):
Tested various swaps + approvals on:
- mainnet
- polygon
- gravity
- apechain
- base
- token to token
- eth to token
- token to eth
- wraps
- crosschain

Also tested:
- swapping into a popular* RST
- swapping into a brand new RST
- swapping RST for RST

src/core/raps/utils.ts
derHowie: done

PR #1871, opened 2025-04-04
fix dep warnings and migrate from createStore -> createRainbowStore

walmat:
Fixes bunch of deprecation warnings that were popping up from Zustand bump. I also noticed that Zustand introduced a breaking change in 4.5.5 that didn't persist initial state into storage until you call setState manually, so I bumped us back a version to 4.5.4. Shouldn't have any effect for current prod, but something that will if everything was using createRainbowStore.

Screenshot 2025-04-04 at 7 59 46 PM

## What to test
State migration flow should be tested. But nothing should be effected since it's using the same storageKey mapping.


walmat:
closing in favor of #1873

PR #1873, opened 2025-04-08
Replace `withSelectors` usage with hook derived from createRainbowStore

walmat:
Converts a lot of our existing Zustand stores to use create() from Zustand instead of the vanilla version. It also removed the need for the withSelectors wrapper which required us to sync two separate versions of the stores. This PR should (in theory) cut memory of the BX in half, but still need to profile it a bit.

n/a

## What to test
full regression testing needed.


package.json
```diff
@@ -167,7 +167,7 @@
     "viem": "2.21.55",
     "wagmi": "2.8.1",
     "word-wrap": "1.2.4",
-    "zustand": "4.5.5"
+    "zustand": "4.5.4"
```
walmat, line 170: reason here: https://github.com/pmndrs/zustand/discussions/2763

BrodyHughes (approved):
Covered full regression testing checklist here: https://www.notion.so/rainbowdotme/QA-Critical-Path-Checklist-for-BX-98fea5b861254eeabec0b09e4dc7ffef

No issues found 👀

DanielSinclair (approved):

PR #1880, opened 2025-04-23
fix message decoding

BrodyHughes:
Fixes BX-1776 (CAAAW!!!! 🦅🇺🇸🎆)
Figma link (if any):

## What changed (plus any additional context for devs):

Fixes an issue where hex-encoded messages in personal_sign requests were being incorrectly identified as Ethereum addresses, causing the message and address parameters to be swapped. This resulted in unreadable text being displayed to users.

Before:
Screenshot 2025-04-23 at 3 31 12 PM

After:
Screenshot 2025-04-23 at 3 55 25 PM

### DIFF
--- a/src/core/utils/signMessages.tsx
+++ b/src/core/utils/signMessages.tsx
@@ -18,14 +18,30 @@ export const getSigningRequestDisplayDetails = (
   try {
     switch (payload.method) {
       case 'personal_sign': {
-        let message = payload?.params?.[0] as string;
-        let address = payload?.params?.[1] as Address;
-        // While this is technically incorrect
-        // we'll support anyways for compatibility purposes
-        // since other wallets do it
-        if (isAddress(message)) {
-          message = payload?.params?.[1] as string;
-          address = payload?.params?.[0] as Address;
+        // Get both parameters from the request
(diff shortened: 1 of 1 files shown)


src/core/utils/signMessages.tsx
```diff
@@ -18,14 +18,25 @@ export const getSigningRequestDisplayDetails = (
   try {
     switch (payload.method) {
       case 'personal_sign': {
-        let message = payload?.params?.[0] as string;
-        let address = payload?.params?.[1] as Address;
-        // While this is technically incorrect
-        // we'll support anyways for compatibility purposes
-        // since other wallets do it
-        if (isAddress(message)) {
```
DanielSinclair, line 26: It took me a lil bit to figure out why this wouldn't work correctly. It turns out `Sign in to Fileverse` is actually a 42 character hex string that does conform to an address. I think a more simple modification here would be to only do the param flip when message is an address and address is not an address (so checking both params rather than just the first one)

DanielSinclair: Would just use isAddress for that (rather than recreating)

walmat (approved):

DanielSinclair (approved):

PR #1888, opened 2025-04-30
Insufficient gas warning not displayed when no nativeAsset

BrodyHughes:
## Problem:
Users who only own non-native tokens on a chain (like wDEGEN but no DEGEN) weren't seeing gas warnings when attempting to sign txns only a blank sign transaction screen with a cancel button. This occurred because our UI relied solely on useUserNativeAsset which returns undefined when the native token isn't in the wallet.

## Solution:

- Added useNativeAsset hook to the InsuficientGasFunds component to always fetch the chain's native asset metadata
- Kept useUserNativeAsset for checking user balances
- Modified UI logic to use chain metadata for display purposes while still checking user balances for conditional messages
Updated button action handlers to use the appropriate asset reference

## UI Comparison

| Before | After |
|--------|-------|
| UI Before | UI After |

### DIFF
--- a/src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
+++ b/src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
@@ -439,13 +439,13 @@ function InsuficientGasFunds({
       <Inline alignVertical="center" space="12px">
         <ChainBadge chainId={chainId} size={16} />
         <Text size="14pt" weight="bold">
-          {nativeAsset && +nativeAsset?.balance.amount > 0
+          {+(nativeAsset?.balance?.amount ?? '0') > 0
             ? i18n.t('approve_request.insufficient_gas_funds', { token })
             : i18n.t('approve_request.no_gas_funds', { token })}
         </Text>
       </Inline>
       <Text size="12pt" weight="medium" color="labelQuaternary">
-        {nativeAsset && +nativeAsset?.balance.amount > 0
+        {+(nativeAsset?.balance?.amount ?? '0') > 0
           ? i18n.t('approve_request.insufficient_gas_funds_description', {
               token,
             })
--- a/src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
+++ b/src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
@@ -437,13 +437,13 @@ function InsuficientGasFunds({
       <Inline alignVertical="center" space="12px">
         <ChainBadge chainId={chainId} size={16} />
         <Text size="14pt" weight="bold">
-          {nativeAsset
+          {nativeAsset && +nativeAsset.balance.amount > 0
             ? i18n.t('approve_request.insufficient_gas_funds', { token })
             : i18n.t('approve_request.no_gas_funds', { token })}
         </Text>
       </Inline>
       <Text size="12pt" weight="medium" color="labelQuaternary">
-        {nativeAsset
+        {nativeAsset && +nativeAsset.balance.amount > 0
           ? i18n.t('approve_request.insufficient_gas_funds_description', {
               token,
             })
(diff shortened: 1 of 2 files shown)


src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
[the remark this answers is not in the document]
BrodyHughes: i have no clue

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
```diff
@@ -437,13 +439,13 @@ function InsuficientGasFunds({
...
         <Text size="14pt" weight="bold">
-          {+nativeAsset.balance.amount > 0
+          {nativeAsset && +nativeAsset?.balance.amount > 0
             ? i18n.t('approve_request.insufficient_gas_funds', { token })
             : i18n.t('approve_request.no_gas_funds', { token })}
         </Text>
       </Inline>
       <Text size="12pt" weight="medium" color="labelQuaternary">
-        {+nativeAsset.balance.amount > 0
+        {nativeAsset && +nativeAsset?.balance.amount > 0
```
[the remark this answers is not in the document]
BrodyHughes: i like yours better

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
```diff
@@ -437,13 +439,13 @@ function InsuficientGasFunds({
       <Inline alignVertical="center" space="12px">
         <ChainBadge chainId={chainId} size={16} />
         <Text size="14pt" weight="bold">
-          {+nativeAsset.balance.amount > 0
+          {+(nativeAsset?.balance?.amount ?? '0') > 0
```
DanielSinclair, line 442: This would be an invalid comparison between a string and number. The `nativeAsset?` should be enough to suffice for the check (no gas when the balance is equal to 0 or nativeAsset is null)

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
```diff
@@ -437,13 +439,13 @@ function InsuficientGasFunds({
...
             ? i18n.t('approve_request.insufficient_gas_funds', { token })
             : i18n.t('approve_request.no_gas_funds', { token })}
         </Text>
       </Inline>
       <Text size="12pt" weight="medium" color="labelQuaternary">
-        {+nativeAsset.balance.amount > 0
+        {+(nativeAsset?.balance?.amount ?? '0') > 0
```
DanielSinclair, line 448: Same as above for this one

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
```diff
@@ -461,13 +463,15 @@ function InsuficientGasFunds({
...
             onClick={() => {
-              setSelectedToken(nativeAsset);
-              navigate(ROUTES.BRIDGE, { replace: true });
-              onRejectRequest({ preventWindowClose: true });
-              triggerToast({
-                title: i18n.t('approve_request.request_rejected'),
-                description: i18n.t('approve_request.bridge_and_try_again'),
-              });
+              if (nativeAsset) {
+                setSelectedToken(nativeAsset);
```
DanielSinclair, line 467: Was there a reason we need this check? The if wrapper here would just mean that the button does nothing when clicked, which would be an awkward state. `hasBridgeableBalance` should be doing the work here to ensure that nativeAsset is not null

DanielSinclair (changes requested):
Added a few comments

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
BrodyHughes: i think with type coercion theres no issue in the comparison (?). I can look into the other check though in the morning.

src/entries/popup/pages/messages/SendTransaction/SendTransactionsInfo.tsx
BrodyHughes: it was causing a ts error bc nativeAsset could be undefined

DanielSinclair (approved):
My recommendations were a bit off; made one more tweak. lgtm!

PR #1917, opened 2025-07-13
fix: contract validation for smart contract sends

DanielSinclair:
BX-1808

## Summary
- added a token contract check using known asset list
- update explainer text for token contract warning
-  added unit tests for u`seSendValidation` hook

## Testing
- test a send to a smart contract and an ERC-20 contract
- test a send using an EOA address or contact/ENS
- you should only see the token contract warning for known ERC-20s

------
https://chatgpt.com/codex/tasks/task_e_6872c3ba3c408325be8b8bfe4749f9ca

---

## PR-Codex overview
This PR focuses on renaming and updating the handling of addresses related to smart contracts to token contracts, along with enhancing validation logic for sending assets.

### Detailed summary
- Renamed `to_smart_contract` to `to_token_contract` in `static/json/languages/en_US.json`.
- Updated descriptions related to token contracts.
- Changed variable names from `toAddressIsSmartContract` to `toAddressIsTokenContract` in `src/entries/popup/pages/send/index.tsx`.
- Modified the `showToContractExplainer` function to `showToTokenContractExplainer`.
- Adjusted validation logic in `useSendValidations` to identify token contracts.
- Added test cases for validating sending to token contracts in `src/entries/popup/hooks/send/useSendValidations.test.ts`.

### DIFF
--- a/src/entries/popup/hooks/send/useSendValidations.test.ts
+++ b/src/entries/popup/hooks/send/useSendValidations.test.ts
@@ -1,0 +1,72 @@
+import { renderHook } from '@testing-library/react';
+import { Address } from 'viem';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+import { useUserAssets } from '~/core/resources/assets';
+import { useCurrentAddressStore, useCurrentCurrencyStore } from '~/core/state';
+import { ChainId } from '~/core/types/chains';
+import {
+  DAI_MAINNET_ASSET,
+  TEST_ADDRESS_1,
+  USDC_MAINNET_ASSET,
+} from '~/test/utils';
+
+import { useSendValidations } from './useSendValidations';
+vi.mock('../useUserNativeAsset', () => ({
+  useUserNativeAsset: vi.fn().mockReturnValue({
+    nativeAsset: { balance: { amount: '0' } },
+  }),
+}));
+
+vi.mock('~/core/resources/assets', () => ({
+  useUserAssets: vi.fn(),
+}));
+
+const USER_ASSETS = {
+  [ChainId.mainnet]: {
+    [DAI_MAINNET_ASSET.uniqueId]: DAI_MAINNET_ASSET,
+    [USDC_MAINNET_ASSET.uniqueId]: USDC_MAINNET_ASSET,
+  },
+};
+
+beforeEach(() => {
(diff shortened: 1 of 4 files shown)


DanielSinclair:
@janek26 Added you for review! This should unblock e2e failures going forward.

After research, I leaned against the ERC-165 check because the majority of ERC-20s don't support 165 and escape that route. To be exhaustive, we'd need to do the following checks against the provider, and it seems impractical. I'm just checking against the user's token list currently.

```md
Step 1 – get the byte‑code
• Call eth_getCode(addr).
• If the result is exactly “0x” (length 0) the address is an EOA and you can stop; otherwise continue.

Step 2 – follow proxies early
• If the runtime code is very small (≈100–200 bytes) or matches a known EIP‑1967/UUPS/clone pattern, read the implementation pointer from storage (EIP‑1967 slot 0x3603… etc.).
• Replace addr with the implementation address for all later checks.

Step 3 – ERC‑20 probe (fungible)
• staticcall selector 0x18160ddd (totalSupply).
• staticcall selector 0x70a08231 (balanceOf(address)) using any address.
• Both calls must return data (no revert).
• Optionally confirm with decimals() 0x313ce567 or recent Transfer events.
• If this passes, tentatively label as ERC‑20 and skip to Step 6 unless you need to check the other standards first.

Step 4 – ERC‑721 probe (NFT)
• staticcall ownerOf(uint256) 0x6352211e with tokenId 0.
• staticcall balanceOf(address) 0x70a08231 with any address.
• staticcall safeTransferFrom(address,address,uint256) 0x42842e0e (or the four‑arg overload 0xb88d4fde) with dummy values.
• If at least two of these three calls return data, label as ERC‑721.

Step 5 – ERC‑1155 probe (multi‑token)
• staticcall balanceOfBatch(address[],uint256[]) 0x4e1273f4 with empty arrays or minimal data.
• staticcall safeTransferFrom(address,address,uint256,uint256,bytes) 0xf242432a or safeBatchTransferFrom 0x2eb2c2d6 with dummy values.
• If balanceOfBatch plus either transfer call both return data, label as ERC‑1155.

Step 6 – disambiguation and fallback
• If multiple standards appear to match, prefer ERC‑1155 over ERC‑721 over ERC‑20 (because later standards often include earlier functions).
• If none of the probes succeed after following proxies, classify as “other contract”.
```

PR #1918, opened 2025-07-15
Refactor/BX-1810-remove-refraction

janek26:
Fixes BX-1810
Figma link (if any):

## What changed

Remove unused refraction service

## What to test

This change should not have any impact, as the service was not used. A light smoke test was done


DanielSinclair (approved):
Looks good to me! Did a little rebase while testing to revert some of the generated file changes; the design system tokens are generated at yarn setup, so I suspect your SF Symbols slightly differs. You'll just want to skip the other generated file check-ins (lockfile, policy) too unless there's a dependency change.

PR #1923, opened 2025-07-16
feat: unite connection requests

janek26:
Fixes BX-1802

## What changed

If the same tab with the same host sends 2 or more connection requests, we just focus the existing approve window and resolve all requests based on that

## What to test

Try to break the connection flow with any dapp, try to connect multiple times on the same host and tab, and same host in 2 different tabs etc

### DIFF
--- a/src/core/state/requests/index.ts
+++ b/src/core/state/requests/index.ts
@@ -1,25 +1,87 @@
 import { createRainbowStore } from '~/core/state/internal/createRainbowStore';
 
-import { ProviderRequestPayload } from '../../transports/providerRequestTransport';
+import type { ProviderRequestPayload } from '../../transports/providerRequestTransport';
+
+import { getSenderHost, getTabIdString } from './utils';
+
+type PromiseResolver = {
+  resolve: (value: object) => void;
+  reject: (error: Error) => void;
+};
 
 export interface PendingRequestsStore {
   pendingRequests: ProviderRequestPayload[];
-  addPendingRequest: (request: ProviderRequestPayload) => void;
+  connectionPromiseResolvers: Map<string, PromiseResolver[]>;
+  addPendingRequest: (request: ProviderRequestPayload) => boolean;
   removePendingRequest: (id: number) => void;
+  addConnectionResolver: (host: string, resolver: PromiseResolver) => void;
+  resolveConnectionRequests: (
+    host: string,
+    result: unknown,
+    isError?: boolean,
+  ) => void;
 }
 
 export const usePendingRequestStore = createRainbowStore<PendingRequestsStore>(
   (set, get) => ({
     pendingRequests: [],
+    connectionPromiseResolvers: new Map(),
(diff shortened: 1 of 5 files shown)


src/entries/background/handlers/handleProviderRequest.ts
```diff
@@ -66,62 +67,106 @@ const focusOnWindow = (windowId: number) => {
...
-  // Add pending request to global background state.
-  addPendingRequest(request);
+  const { addPendingRequest, addConnectionResolver } =
+    usePendingRequestStore.getState();
+
+  // Check if this is a duplicate connection request
+  const isRequestAdded = addPendingRequest(request);
 
-  let ready = isInitialized();
-  while (!ready) {
```
DanielSinclair, line 103: IIRC ready was related to inpage script bootup delays

src/core/state/requests/index.ts
```diff
@@ -1,26 +1,88 @@
...
       set({ pendingRequests: pendingRequests.concat([newRequest]) });
+      return true;
     },
     removePendingRequest: (id) => {
       const pendingRequests = get().pendingRequests;
       set({
         pendingRequests: pendingRequests.filter((request) => request.id !== id),
       });
     },
+    addConnectionResolver: (host, resolver) => {
```
DanielSinclair, line 62: Can you add some notes about the resolver paradigm used here? Trying to wrap my head around the goal.

src/entries/background/handlers/handleProviderRequest.ts
```diff
@@ -66,62 +67,106 @@ const focusOnWindow = (windowId: number) => {
...
+      reject: (error: Error) => void;
+    },
+  ) => void,
+): Promise<object> => {
+  return new Promise((resolve, reject) => {
+    addConnectionResolver(host, { resolve, reject });
+  });
+};
+
+const resolveConnectionRequestsIfNeeded = (
```
DanielSinclair, line 102: Is there any duplicative implementation between this concept and `handleTabAndWindowUpdates`? We attach to the window/tab listeners there for things like expiry

DanielSinclair (changes requested):
Main issue I noticed was a few new failing unit tests. You should be able to run those locally with `yarn test` to debug. Trying to understand more about the resolver concept here as well

Here is the failing test logs: https://github.com/rainbow-me/browser-extension/actions/runs/16313515910/job/46074063502?pr=1923

src/core/state/requests/index.ts
janek26: added some in code comments :) Basically it's needed to allow multiple connect requests to resolve when one dialog gets accepted/declined

DanielSinclair:
@janek26 Code lgtm, but the unit tests related to the provider are failing here. That suite is actually working; you'll just need to run `yarn test`.  I'll do some aggressive testing and can get this merged tomorrow if you're able to sort that out

PR #1924, opened 2025-07-16
fix: fallback to asset instead of change if needed

janek26:
Fixes BX-1795

## What changed

Smarter way of determining change of a transaction relative to the user
For some events, like airdrops, the backend return a response that looks like this:
```
{
  ...
  direction: "out",
  changes [{
    ...
    direction: "in"
  }]
}
```

I guess this makes sense as the transaction was a claim transaction (sending a tx = out) but the effect (balance change) is that the token balance increases. The activities mostly reflect what happens to your balance, so this PR makes sure we use the direction of balance changes over the direction of the transaction. This fixed the mentioned issue

Bildschirmfoto 2025-07-16 um 13 07 31

## What to test

Please test receiving, sending and bridging assets on different chains, and make sure the direction shown in rainbow activities matches the actual activity

---

## PR-Codex overview
This PR focuses on enhancing transaction handling and asset validation in a React application. It modifies ESLint rules, introduces a new function for asset validation, and updates transaction direction logic, improving the accuracy and clarity of asset-related operations.

### Detailed summary
- Updated `.eslintrc.js` to turn off specific rules.
- Added `isParsedUserAsset` function to validate assets.
- Modified `getDirection` function to include transaction changes.
- Updated handling of transaction properties in `ActivityValue.tsx`.
- Improved asset extraction and validation logic in transaction processing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
   extends: ['rainbow', 'plugin:prettier/recommended'],
   root: true,
   rules: {
+    'no-nested-ternary': 'off',
     'import/no-default-export': 'off',
     'react/react-in-jsx-scope': 'off',
     'import/order': [
--- a/src/core/utils/assets.ts
+++ b/src/core/utils/assets.ts
@@ -232,6 +232,12 @@ export function parseUserAssetBalances({
   };
 }
 
+export function isParsedUserAsset(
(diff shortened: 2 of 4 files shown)


DanielSinclair (approved):
Functionality looks good to me!

DanielSinclair:
Was there any logic introduced here or in this display type mapping that we might want to push to Backend? They're currently refactoring this indexing service if so!

PR #1928, opened 2025-07-17
fix: correct perceived finality API query parameter format

janek26:
Update staleBalances to use `tokens` parameter with comma-separated values instead of multiple `token` parameters. This aligns with the backend API expectations for the perceived finality system.

Changes:
- Change parameter name from `token` to `tokens`
- Format multiple tokens as comma-separated list
- Update tests to match new format

---

## PR-Codex overview
This PR refactors the `getStaleBalancesQueryParam` function to return a list of tokens instead of a single query string fragment. It also updates the corresponding tests to reflect this change in the expected output format.

### Detailed summary
- Changed `getStaleBalancesQueryParam` to build a `tokenList` instead of a `queryStringFragment`.
- Updated return value to `&tokens=${tokenList.join(',')}`.
- Modified tests to expect `&tokens=` format instead of `&token=` for query parameters.


DanielSinclair (approved):
Nice!

PR #1948, opened 2025-07-23
fix: custom networks

janek26:
Fixes BX-1787

## What changed

Fixed bug which made it impossible to add custom networks by dapps

## What to test

Use chainlist to add custom networks and confirm they get added. Also confirm reject and the case where a network already exists get handled


src/core/state/networks/networks.ts
```diff
@@ -198,10 +198,10 @@ function createSelector<T>(
 function createParameterizedSelector<T, Args extends unknown[]>(
   selectorFn: (params: {
     networks: Networks;
-    userPreferences: Record<number, ChainPreferences>;
+    userPreferences: Record<number, ChainPreferences | undefined>;
     chainOrder: number[];
     enabledChainIds: Set<number>;
-    mergedChainData: Record<number, TransformedChain>;
+    mergedChainData: Record<number, TransformedChain | undefined>;
```
DanielSinclair, line 204: Was there a reason we needed undefined here?

src/entries/background/handlers/handleProviderRequest.ts
```diff
@@ -261,8 +261,7 @@ export const handleProviderRequest = ({
     removeAppSession: ({ host }: { host: string }) =>
       useAppSessionsStore.getState().removeAppSession({ host }),
     getChainNativeCurrency: (chainId: number) =>
-      useNetworkStore.getState().getBackendSupportedChains(true)[chainId]
-        ?.nativeCurrency,
+      useNetworkStore.getState().getChain(chainId)?.nativeCurrency,
```
DanielSinclair, line 264: Nice find 👌

src/entries/popup/pages/messages/AddEthereumChain/index.tsx
```diff
@@ -79,6 +78,13 @@ export const AddEthereumChain = ({
         dappDomain: dappMetadata?.appHost || '',
         dappName: dappMetadata?.appName,
       });
+
+      // wait 1.2s for storages to sync (persist throttle on network store is 1s)
+      await new Promise((resolve) => {
+        setTimeout(resolve, 1200);
+      });
+
+      approveRequest(true);
```
DanielSinclair, line 87: Would keep this above the analytics event in case that is ever blocking

src/core/state/networks/networks.ts
janek26: in a record, undefined is always a possibility :)
bth I'm not sure why typescript does not do this by default.
There's no Record (object, map) where each key is defined, unless you work with getters

src/entries/popup/pages/messages/AddEthereumChain/index.tsx
janek26: I updated the analytics to make it non-blocking

I think keeping it infront of the timeout is a good idea as it gives more time for the network request to succeed
approveRequest() will close the popup and the event may never gets delivered

PR #1949, opened 2025-07-23
fix: invalidate gas estimate on chain switch

janek26:
Fixes BX-1819

## What changed

Gas has invalidate when sswitching chains, otherwise gas estimates from one chain may be used on another

## What to test

The issue is hard to reproduce, if you find a way to reprod BX-1819 on a build before this PR gets merged, let me know

### DIFF
--- a/src/core/state/gas/index.ts
+++ b/src/core/state/gas/index.ts
@@ -24,6 +24,7 @@ export interface GasStore {
     gasFeeParamsBySpeed: GasFeeParamsBySpeed | GasFeeLegacyParamsBySpeed;
   }) => void;
   clearCustomGasModified: () => void;
+  clearGasData: () => void;
 }
 
 export const useGasStore = createRainbowStore<GasStore>(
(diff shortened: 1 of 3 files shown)


DanielSinclair (changes requested):
I think this issue is a bit deeper. This store is generally for the "custom gas" feature where the user can create custom gas, or select their default speed from the drop-down menu on transaction interfaces. Whereas live gas values are fetched from Meteorology and assembled upstream in `useGas` or the derivatives like `useSendGas`. Found the diagram here to be helpful: https://deepwiki.com/search/describe-the-overlap-between-u_0f96509d-05a8-4627-acf2-af2647c840da

I think the general direction here is right, but I think it would be better to focus on a repro and refactoring the gas fetching to reflect provider chain changes in `useGas`. The user report seems to indicate that this happened without touching gas settings via the dapp provider. A little biased against adding another temperamental syncStore fn as well; hopefully we can do away with that entirely!

janek26:
I would love to refactor everything gas, do you think that would be okay? I think it's a great task, as it's quite isolated

PR #1954, opened 2025-07-24
Feature/BX-1830-handle-loading-for-out-of-funds

janek26:
Fixes BX-1830

## What changed

Added loading indicators if we cant determine if user has enough eth to pay for transaction yet or not

## What to test

Reprod in this ticket: https://linear.app/rainbow/issue/BX-1829/insufficient-eth-error-initially-when-reconnecting-to-dapp-after

---

## PR-Codex overview
This PR enhances the `useHasEnoughGas` hook and its usage across various components to handle loading states more effectively. It introduces loading indicators and adjusts the logic for checking if users have enough gas, improving user experience during balance checks.

### Detailed summary
- Updated `useHasEnoughGas` to return `isLoading` alongside `hasEnough`.
- Added loading state checks in components using `useHasEnoughGas`.
- Introduced a `Skeleton` component for loading indicators in `SendTransactionInfo` and `WalletNativeBalance`.
- Adjusted conditional rendering logic to handle loading states in UI components.
- Modified `useUserNativeAsset` to return `isLoading` status for better asset fetching management.

### DIFF
--- a/src/entries/popup/hooks/approveAppRequest/useApproveAppRequestValidations.ts
+++ b/src/entries/popup/hooks/approveAppRequest/useApproveAppRequestValidations.ts
@@ -21,7 +21,8 @@ export const useApproveAppRequestValidations = ({
   const { connectedToHardhat, connectedToHardhatOp } =
     useConnectedToHardhatStore();
 
-  const enoughNativeAssetForGas = useHasEnoughGas(session);
+  const { hasEnough: enoughNativeAssetForGas, isLoading: isGasLoading } =
+    useHasEnoughGas(session);
 
   const buttonLabel = useMemo(() => {
     const activeChainId = chainIdToUse(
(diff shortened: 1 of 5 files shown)


DanielSinclair (changes requested):
Oddly can't seem to trigger the tx loading skeleton with DevTools throttling, but the `isLoading` state changes seem to have improved this build out a lot. I did see some small sizing regressions for `SendTransactionsInfo` height. I wonder if `motion.div` is the culprit; maybe we can just abandon that wrapping div? Some of the animations on this flow are already a bit jarring to be honest.

| Before  | After |
| ------------- | ------------- |
| Image | Image  |
| Image  | Image |

janek26:
thanks for the regression test, fixed :) The skeleton will only show if your gas fee token balance is unknown

BrodyHughes (approved):
couldnt repro original issue but the behavior seems improved to me. tested on a few dapps.

PR #1958, opened 2025-07-25
fix: handle chart errors

janek26:
Fixes BX-1833

## What changed

The underlying issue is not fixed, but also not reproducable and it looks like it was temporary. My guess is due to missing data or malformed data from backend.
This PR makes sure only the chart crashes next time, and the rest stays functional

https://github.com/user-attachments/assets/11c7b458-22c2-49e6-8d4e-f07147f25519


DanielSinclair (approved):
lgtm

PR #1998, opened 2025-08-19
refactor: standardize browser config

DanielSinclair:
Fixes BX-####
Figma link (if any):


BrodyHughes (approved):
Only concern is maybe clarifying what directory to install this in so user doesnt install in the bx directory

```
    console.error('To fix this, install the correct version:');
    console.error(`npx @puppeteer/browsers install chrome@${expectedVersion}`);
    ```

PR #2021, opened 2025-08-28
test: add E2E tests for custom networks and RPCs

BrodyHughes:
Fixes BX-1853

- Add initial state check to capture baseline network/RPC counts
- Implement test for adding custom RPC to mainnet via dapp
- Implement test for adding custom network via dapp
- Add verification after disconnect to ensure persistence
- Follow proper test ordering with disconnect as final step


DanielSinclair (approved):
Nice!

PR #2033, opened 2025-09-06
fix: disable nft fetch in search hooks based on remote config

DanielSinclair:
Fixes BX-####
Figma link (if any):


DanielSinclair:
Merging this because it's fairly small change to fix fetch errors, and also fixes a CI lint error introduced to master

PR #2047, opened 2025-09-17
feat: add more breadcrumbs to internal builds

janek26:

- Exported `INTERNAL_BUILD` and `IS_TESTING` constants from the Sentry module to make them accessible in other parts of the application
- Added Sentry breadcrumbs for ORPC calls in the `sentryMiddleware` to improve debugging capabilities in internal builds and testing environments
- Added Sentry breadcrumbs for wallet address access to track wallet-related operations during development and testing

## What to test

- Verify that Sentry breadcrumbs are properly added for ORPC calls in internal builds and testing environments
- Confirm that wallet address operations are correctly tracked with breadcrumbs
- Ensure that breadcrumbs are only added when `INTERNAL_BUILD` or `IS_TESTING` is true
- Check that the existing error handling in the Sentry middleware still works as expected

### DIFF
--- a/src/entries/background/procedures/popup/index.ts
+++ b/src/entries/background/procedures/popup/index.ts
@@ -1,6 +1,8 @@
 import { os } from '@orpc/server';
 import { RPCHandler } from '@orpc/server/message-port';
+import * as Sentry from '@sentry/react';
 
+import { INTERNAL_BUILD, IS_TESTING } from '~/core/sentry';
 import { RainbowError, logger } from '~/logger';
 
 import { healthRouter } from './health';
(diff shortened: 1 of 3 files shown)


src/entries/background/procedures/popup/index.ts
```diff
@@ -9,7 +11,16 @@ import { stateRouter } from './state';
...
-const sentryMiddleware = os.middleware(async ({ next }) => {
+const sentryMiddleware = os.middleware(async ({ next, path }) => {
+  if (INTERNAL_BUILD || IS_TESTING) {
+    Sentry.addBreadcrumb({
+      message: `ORPC: ${path.join('/')}`,
+      data: {
+        path: path.join('/'),
+        timestamp: new Date().toISOString(),
+      },
+    });
```
DanielSinclair, line 22: Any reason we don't want to expose this for all releases?

janek26: I think it's too much data, and it will be pretty repetitive
lets see, we can always keep it

PR #2051, opened 2025-09-19
fix: internal keychain tracing

janek26:
Fixes BX-####
Figma link (if any):

### DIFF
--- a/src/core/keychain/KeychainManager.ts
+++ b/src/core/keychain/KeychainManager.ts
@@ -7,10 +7,12 @@ import {
   encryptWithKey,
   importKey,
 } from '@metamask/browser-passworder';
+import * as Sentry from '@sentry/react';
 import { Address } from 'viem';
 
 import { RainbowError, logger } from '~/logger';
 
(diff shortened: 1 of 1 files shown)


src/core/keychain/KeychainManager.ts
```diff
@@ -7,10 +7,12 @@ import {
   encryptWithKey,
   importKey,
 } from '@metamask/browser-passworder';
+import * as Sentry from '@sentry/react';
 import { Address } from 'viem';
 
 import { RainbowError, logger } from '~/logger';
 
+import { INTERNAL_BUILD, IS_TESTING } from '../sentry';
```
DanielSinclair, line 15: @claude To reduce the extra import, can you just use:
```
export const INTERNAL_BUILD = process.env.INTERNAL_BUILD === 'true';
export const IS_TESTING = process.env.IS_TESTING === 'true';
```

DanielSinclair (approved):

src/core/keychain/KeychainManager.ts
DanielSinclair: @claude Fix the lint error

PR #2056, opened 2025-09-22
feat: log client orpc errors to sentry

janek26:
Fixes BX-####
Figma link (if any):


src/entries/popup/handlers/background.ts
```diff
@@ -1,23 +1,34 @@
...
 function createPort() {
   return chrome.runtime.connect({ name: POPUP_PORT_NAME });
 }
+function createClient(port: chrome.runtime.Port): RouterClient<PopupRouter> {
+  return createORPCClient(
+    new RPCLink({
+      port,
+      clientInterceptors: [
+        onError((e) => {
+          logger.error(new RainbowError('ORPC client error'), { cause: e });
```
DanielSinclair, line 22: @claude Make sure we're passing `cause` to RainbowError after the message

DanielSinclair (commented):
@claude Make sure we're passing cause to RainbowError after the message

PR #2067, opened 2025-09-26
feat: enhance date formatting utilities

janek26:
- Introduced `formatExactDateTime` for precise date-time formatting with locale support.
- Updated `formatRelativeDate` to improve readability and added "Just now" for recent timestamps.
- Refactored components to utilize the new date formatting functions, enhancing the user experience in displaying transaction times and activity details.

### DIFF
--- a/src/core/utils/formatDate.ts
+++ b/src/core/utils/formatDate.ts
@@ -1,25 +1,48 @@
 import { i18n } from '../languages';
 
-export function formatDate(date: string | number | Date) {
+export function formatExactDateTime(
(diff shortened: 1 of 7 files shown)


DanielSinclair (commented):
Small issues:
- Seems like the localized string is still surfacing 24h time format for en_US (in the tooltip)
- Can't seem to select the text with cursor for the label on Activity Details as well as the`Block` value below it
- To mirror app we could do `Sept 29th, 2025 at 7:39 PM` in the main label; don't mind keeping the tooltip

PR #2262, opened 2026-04-07
fix(send): gate in-app and dApp approval on native balance for value + max fee

janek26:
Users could advance send / approval flows even when they did not hold enough of the chain’s native asset to cover the **transfer and** the worst-case network fee. That showed up most clearly when a dApp proposed a transaction with **native `value`**: the approval path only compared **gas** to balance, so the flow could treat “gas only” as enough and only fail at broadcast, often with a generic error.

## What changed

We now **gate the UI** on the same rule where it matters: **native balance must cover `value` + `selectedGas.gasFee.amount`** (the max fee line from the gas store, including OP-stack L1 security fee where that feed applies). The check is **`bigint`-only**, implemented once in `src/core/send/nativeSendBalance.ts` as `hasEnoughNativeBalanceForSend`.

- **In-app send** — `useSendValidations` uses that helper (native send: amount + gas; ERC-20 send: gas only).
- **DApp `eth_sendTransaction`** — `useHasEnoughGas` takes optional `nativeValueHex` derived from the pending tx (`SendTransaction`) so approval is not gas-only when the tx carries native value.

There is **no** extra balance guard inside `handlers/wallet.sendTransaction`; validation lives in hooks so we don’t duplicate logic at submit.

Fixes BX-2073.

---

## PR-Codex overview
This PR introduces enhancements to handle native asset balance checks during transactions, ensuring that the total amount and gas fees do not exceed the user's balance. It adds new functions and updates existing hooks to utilize these checks for better validation.

### Detailed summary
- Added `insufficient_native_funds_description` to `static/json/languages/en_US.json`.
- Implemented `hasEnoughNativeBalanceForSend` function in `src/core/send/nativeSendBalance.ts` for balance validation.
- Updated `useApproveAppRequestValidations` to accept `nativeValueHex`.
- Modified `SendTransactionActions` to pass `nativeValueHex` to validation hooks.
- Enhanced `useSendValidations` to utilize `hasEnoughNativeBalanceForSend`.
- Updated `useHasEnoughGas` to check if the balance covers both value and gas fees.
- Created tests for `hasEnoughNativeBalanceForSend` in `src/core/send/nativeSendBalance.test.ts`.
- Adjusted `SendTransaction` component to compute and pass `nativeValueHexForBalanceCheck`.

### DIFF
--- a/src/entries/popup/pages/messages/SendTransaction/index.tsx
+++ b/src/entries/popup/pages/messages/SendTransaction/index.tsx
@@ -1,6 +1,7 @@
 import { TransactionRequest } from '@ethersproject/abstract-provider';
+import { BigNumber } from '@ethersproject/bignumber';
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { Address, getAddress } from 'viem';
+import { Address, type Hex, getAddress } from 'viem';
 
 import { analytics } from '~/analytics';
 import { event } from '~/analytics/event';
@@ -120,6 +121,21 @@ export function SendTransaction({
     chainId,
     selectedWallet,
   ]);
+
+  /** Hex wei for tx `value` — included in `useHasEnoughGas` so balance covers value + gas. */
+  const nativeValueHexForBalanceCheck = useMemo((): Hex | undefined => {
+    if (isSendCalls || !transactionRequest) return undefined;
+    const raw = transactionRequest.value;
+    if (raw === undefined || raw === null) return '0x0';
+    try {
+      return BigNumber.from(raw).toHexString() as Hex;
+    } catch (e) {
+      logger.info('SendTransaction: invalid tx value for balance check', {
+        message: (e as Error)?.message,
+      });
+      return '0x0';
+    }
+  }, [isSendCalls, transactionRequest]);
 
   const simulationTransaction = useMemo((): SimulateTransactionInput | null => {
     if (!sendCallsFlowReady || !transactionRequest) return null;
@@ -416,6 +432,7 @@ export function SendTransaction({
         dappStatusForUi={effectiveDappStatus}
         simulationResult={simulationResult}
         onRejectRequest={rejectRequest}
+        nativeValueHex={nativeValueHexForBalanceCheck}
       />
       <Stack space="20px" padding="20px">
         <Bleed vertical="4px">
@@ -459,6 +476,7 @@ export function SendTransaction({
           loading={loading || (isSendCalls && envelopePending)}
           dappStatus={effectiveDappStatus}
           signingWithDevice={isSigningWithDevice}
+          nativeValueHex={nativeValueHexForBalanceCheck}
         />
       </Stack>
     </Box>
--- a/src/core/send/nativeSendBalance.test.ts
+++ b/src/core/send/nativeSendBalance.test.ts
@@ -1,0 +1,33 @@
+import { expect, test } from 'vitest';
+
+import { hasEnoughNativeBalanceForSend } from './nativeSendBalance';
+
+test('hasEnoughNativeBalanceForSend is false when balance < value + gas', () => {
+  expect(
+    hasEnoughNativeBalanceForSend({
+      balanceWei: 99n,
+      valueWei: 50n,
+      gasFeeWei: 50n,
+    }),
+  ).toBe(false);
+});
+
+test('hasEnoughNativeBalanceForSend is true when balance equals value + gas', () => {
+  expect(
+    hasEnoughNativeBalanceForSend({
+      balanceWei: 100n,
+      valueWei: 50n,
+      gasFeeWei: 50n,
+    }),
+  ).toBe(true);
+});
+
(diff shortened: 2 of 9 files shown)


janek26:
* **#2262** <a href="https://app.graphite.com/github/pr/rainbow-me/browser-extension/2262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainbow-me/browser-extension/2262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/core/utils/insufficientFundsError.ts
```diff
@@ -0,0 +1,77 @@
+import type { Provider } from '@ethersproject/abstract-provider';
+import { BigNumber } from '@ethersproject/bignumber';
```
DanielSinclair: Anyway we can keep viem bigint compatibility here to keep the momentum in that direction?

src/core/utils/insufficientFundsError.test.ts
DanielSinclair, line 1: Would prefer we move this and the implementation alongside the handler implementation or that directory structure to continue to phase out the core/utils/ pattern

src/entries/popup/pages/messages/SendTransaction/index.tsx
```diff
@@ -317,11 +318,26 @@ export function SendTransaction({
           message: (e as Error)?.message,
         },
       );
-      const extractedError = (e as Error).message.split('[')[0];
-      triggerAlert({
-        text: i18n.t('errors.sending_transaction'),
-        description: extractedError,
-      });
+      if (e instanceof ORPCError && e.code === 'INSUFFICIENT_FUNDS') {
```
DanielSinclair: Confused a bit about how we reach this condition. The UI button should be gated with `useSendValidations ` and `enoughNativeAssetForGas` in Send Review (in-app send flow) and the equivalent `useApproveAppRequestValidations` in ` SendTransactionActions` for dapp sends. Are we escaping that validation somehow?

src/core/utils/insufficientFundsError.ts
```diff
@@ -0,0 +1,77 @@
...
+      return true;
+    }
+  }
+
+  const requiredWei = add(valueWei, maxGasCostWei);
+  const balance = await provider.getBalance(from);
+  const balanceStr = balance.toString();
+
+  return !greaterThan(requiredWei, balanceStr);
+}
```
DanielSinclair: Implementing this ourselves can be a bit of a nighmare and error prone. The L1+L2 fee needs to be considered here: 
> On chains returned by getNeedsL1SecurityFeeChainIds() (for example Base and Optimism), the repo already adds an extra optimismL1SecurityFee into selectedGas.gasFee.amount, but this helper only compares value + gasLimit * maxFee/gasPrice against the balance. A user whose balance covers the L2 execution fee but not the extra L1 fee will pass this new precheck and still fail when the transaction is submitted, so the new INSUFFICIENT_FUNDS path never triggers on those networks.

src/entries/popup/handlers/wallet.ts
```diff
@@ -138,6 +140,20 @@ export const sendTransaction = async (
...
+    gasLimitHex: params.gasLimit,
+    valueHex: params.value,
+    maxFeePerGasHex: params.maxFeePerGas,
+    gasPriceHex: params.gasPrice,
+  });
+  if (!sufficient) {
+    throw new ORPCError('INSUFFICIENT_FUNDS', {
+      message: 'Insufficient native funds',
+    });
+  }
```
DanielSinclair: Would we actually just want to bubble up ethers (eventually viem) errors here instead? Both will send `INSUFFICIENT_FUNDS` error considering the gas pricing https://docs.ethers.org/v5/troubleshooting/errors/#help-INSUFFICIENT_FUNDS

DanielSinclair: The combination of UI block + send-time error handling seems like the better low-code bet. Other wallets handle this with sticky failed transactions; i.e. you can top-up and resend the same transaction from the transactions list similar to speed-up.

DanielSinclair (commented):
This is viable, but I am debating whether it is the right solution; relying on wallet client internals for catching this error and just ensuring the user has a retry path seems like the better bet IMO. If the user is somehow escaping the UI handling already, then that transaction could just as well fail onchain for other reasons.

src/core/utils/insufficientFundsError.ts
janek26: Fixed by basing the check on _`selectedGas.gasFee.amount`_ (includes L1 security fee on OP stack via existing gas code), not _`gasLimit * maxFee`_.

PR #2269, opened 2026-04-17
fix: only poll 5792 transactions in background

i1skn:
Fixes: https://linear.app/rainbow/issue/BX-2160/investigate-bx-traffic-spike-on-gettransactionbyhash

## Why?

The pending transaction polling moved into the background service worker in `v1.6.3` so `wallet_getCallsStatus` could keep working after the popup closed. That also caused regular pending transactions to be polled continuously in the background, which drove a large increase in `GetTransactionByHash` traffic even though those transactions only need to be refreshed while the popup is open.

This change narrows background polling back to the flows that actually need it. Normal pending transactions continue to get the same responsive polling while the popup is open, and EIP-5792 batch transactions still keep their status up to date after the popup closes.

## Approach

The background watcher now derives the pollable pending transaction set from popup connection state and passes that snapshot into the polling worker. When the popup is open, it polls all pending transactions; when it is closed, it polls only transactions tied to active EIP-5792 batches.

## Testing

See the video I've attached. There I mock 404 for `GetTransactionByHash` so it's visible that in batch flow it does poll in the background. In the non batch flow (revoke.cash) it only call `GetTransactionByHash` after I've opened popup (it stops polling because `ListTransactions` is firing at the same time and it also have updated status for the transaction).[Screen Recording 2026-04-17 at 14.00.24.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/92a0615f-5ac5-4efb-a9ce-0bcc8a0b2e5b.mov)

- Submit a regular transaction, (usually the popup is closed automatically before it confirms), then reopen the wallet later and verify that the transaction is still shown as in progress at first but quickly resolves to its actual state because background polling was off.
- Execute a 5792 `wallet_sendCalls` batch (for example, on https://jumper.xyz/), then reopen the wallet later and verify that the transaction has already resolved to its actual state because background polling was on.

---

> [!NOTE]
> **Medium Risk**
> Changes background polling scope and scheduling based on popup connection state; mistakes could cause pending transactions to stop updating or increase polling frequency unexpectedly. Also optimizes batch status updates by sharing a single receipt fetch across batches, which could affect batch resolution if receipt handling is wrong.
> 
> **Overview**
> Reduces background service-worker RPC load by **polling only EIP-5792 batch-related pending transactions when the popup is closed**, while continuing to poll *all* pending transactions when the popup is open.
> 
> Refactors the polling pipeline so `handleWatchPendingTransactions` computes the pollable pending-tx snapshot (and separate initial-poll state for all-vs-batch-only) and passes it into `watchPendingTransactions`, and adds popup open/close lifecycle notifications to reschedule polling immediately on popup state changes.
> 
> Separately, `updateBatchesForMinedTx` now fetches a transaction receipt once per mined tx and reuses it across all matching batches to avoid redundant receipt RPC calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc11247151a6e028008973a40fe0f80a13b41836. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/entries/background/handlers/handleWatchPendingTransactions.ts
+++ b/src/entries/background/handlers/handleWatchPendingTransactions.ts
@@ -1,20 +1,62 @@
+import type { Address } from 'viem';
+
 import { usePendingTransactionsStore } from '~/core/state';
+import { getBatchKeysForNonce } from '~/core/state/batches/updateBatchStatus';
 import { getMinPollingIntervalForPendingTxs } from '~/core/state/networks/timing';
 import { watchPendingTransactions } from '~/core/state/transactions/pendingTransactions/watchPendingTransactions';
+import type { RainbowTransaction } from '~/core/types/transactions';
 import { RainbowError, logger } from '~/logger';
+
+import {
+  addPopupPortLifecycleListener,
+  isPopupOpen,
+} from '../procedures/popup/popupPortManager';
 
 let nextPollTimeout: ReturnType<typeof setTimeout> | null = null;
 let nextPollAt = 0;
-let initialPollDone = false;
+let initialAllPendingPollDone = false;
+let initialBatchOnlyPollDone = false;
 let unsubscribe: (() => void) | null = null;
+
+function isBatchTx(tx: RainbowTransaction, address: Address): boolean {
+  return getBatchKeysForNonce(tx.nonce, tx.chainId, address).length > 0;
+}
+
+/**
+ * Returns pending txs that should be polled right now.
+ * - Popup open: all pending txs
+ * - Popup closed: only EIP-5792 batch txs
+ */
+function getPendingTxPollScope(): {
+  pollablePendingTxs: Record<Address, RainbowTransaction[]>;
+  isPollingAllPendingTxs: boolean;
+} {
+  const { pendingTransactions } = usePendingTransactionsStore.getState();
+  const isPollingAllPendingTxs = isPopupOpen();
+
+  if (isPollingAllPendingTxs) {
+    return { pollablePendingTxs: pendingTransactions, isPollingAllPendingTxs };
+  }
+
+  const result: Record<string, RainbowTransaction[]> = {};
+
+  for (const [addr, txs] of Object.entries(pendingTransactions)) {
+    const filtered = txs.filter((tx) => isBatchTx(tx, addr as Address));
+    if (filtered.length > 0) {
+      result[addr] = filtered;
+    }
+  }
+
+  return { pollablePendingTxs: result, isPollingAllPendingTxs };
+}
 
 function scheduleNextPoll(): void {
   if (nextPollTimeout) {
     clearTimeout(nextPollTimeout);
   }
 
-  const { pendingTransactions } = usePendingTransactionsStore.getState();
-  const intervalMs = getMinPollingIntervalForPendingTxs(pendingTransactions);
+  const { pollablePendingTxs } = getPendingTxPollScope();
+  const intervalMs = getMinPollingIntervalForPendingTxs(pollablePendingTxs);
 
   nextPollAt = Date.now() + intervalMs;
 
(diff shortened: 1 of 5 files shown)


i1skn:
* **#2269** <a href="https://app.graphite.com/github/pr/rainbow-me/browser-extension/2269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainbow-me/browser-extension/2269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/entries/background/handlers/handleWatchPendingTransactions.ts
```diff
@@ -1,20 +1,71 @@
...
 let nextPollTimeout: ReturnType<typeof setTimeout> | null = null;
 let nextPollAt = 0;
-let initialPollDone = false;
+let initialAllPendingPollDone = false;
+let initialBatchOnlyPollDone = false;
 let unsubscribe: (() => void) | null = null;
 
+function isPopupOpen() {
+  return getActivePopupPortCount() > 0;
+}
```
maxbbb: nit: probably better placed in `popupPortManager`

src/entries/background/handlers/handleWatchPendingTransactions.ts
```diff
@@ -68,6 +128,16 @@ export async function handleWatchPendingTransactions(): Promise<void> {
...
+
+    // Adjust polling scope when popup opens or closes:
+    // - open: poll all pending txs for responsive UI
+    // - closed: poll only batch txs
+    chrome.runtime.onConnect.addListener((port) => {
+      if (port.name === POPUP_PORT_NAME) {
+        maybeReschedule();
+        port.onDisconnect.addListener(() => maybeReschedule());
+      }
+    });
```
maxbbb: Not sure this is consequential, but given there is a centralized location / pipeline where we're already doing this listening (`startPopupRouter` -> `popupPortManager`), I wonder if it would be better to add a listeners callback set and a way to add to it in `popupPortManager` rather than creating an additional one off listener here.

src/entries/background/handlers/handleWatchPendingTransactions.ts
```diff
@@ -1,20 +1,71 @@
...
+function isBatchTx(tx: RainbowTransaction, address: Address): boolean {
+  if (tx.nonce == null) return false;
+  const { batches } = useBatchStore.getState();
+  return Object.values(batches).some(
+    (batch) =>
+      batch.chainId === tx.chainId &&
+      batch.sender.toLowerCase() === address.toLowerCase() &&
+      batch.nonces.includes(tx.nonce),
+  );
+}
```
maxbbb, line 23: nit: `getBatchKeysForNonce` has essentially the same logic, can we re-use here?

maxbbb (approved):

i1skn:
### Merge activity

* **May 11, 2:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/rainbow-me/browser-extension/2269).
* **May 11, 2:01 PM UTC**: [Graphite](https://app.graphite.com/github/pr/rainbow-me/browser-extension/2269) rebased this pull request as part of a merge.
* **May 11, 2:28 PM UTC**: @i1skn merged this pull request with [Graphite](https://app.graphite.com/github/pr/rainbow-me/browser-extension/2269).

